PR #7, opened 2020-04-28
wrong format for timestamp

atyronesmith:


csibbitt:
Sure; I'll test it later it can't really break more ;)

PR #13, opened 2020-10-15
Build from ubi8 base image

leifmadsen:


build/Dockerfile
```diff
@@ -1,29 +1,21 @@
 # --- build smart gateway ---
-FROM centos:7 AS builder
+FROM registry.access.redhat.com/ubi8 AS builder
```
csibbitt, line 2: :+1:

build/Dockerfile
```diff
@@ -1,29 +1,21 @@
...
 
 LABEL io.k8s.display-name="Smart Gateway" \
       io.k8s.description="A component of the Service Telemetry Framework on the server side that ingests data from AMQP 1.x and provides a metrics scrape endpoint for Prometheus, and forwards events to ElasticSearch" \
-      maintainer="Leif Madsen <leif@redhat.com>"
-
-RUN yum install epel-release -y && \
-        yum update -y --setopt=tsflags=nodocs && \
-        yum clean all && \
-        rm -rf /var/cache/yum
+      maintainer="Leif Madsen <leif+smartgatewayoperator@redhat.com>"
```
csibbitt, line 18: Smart move

build/Dockerfile
```diff
@@ -1,29 +1,21 @@
...
 
 # --- end build, create smart gateway layer ---
-FROM centos:7
+FROM registry.access.redhat.com/ubi8
 
 LABEL io.k8s.display-name="Smart Gateway" \
       io.k8s.description="A component of the Service Telemetry Framework on the server side that ingests data from AMQP 1.x and provides a metrics scrape endpoint for Prometheus, and forwards events to ElasticSearch" \
-      maintainer="Leif Madsen <leif@redhat.com>"
-
-RUN yum install epel-release -y && \
```
csibbitt, line 23: Do you happen to remember what was in epel that is now accounted for in UBI?

csibbitt (approved):

build/Dockerfile
leifmadsen: I don't think we needed epel-release at all to be honest (or it was just a requirement to get golang or git). I suspect it is an artifact of the split between sg-bridge and sg-core, and that epel-release was providing the libraries for QPID (which is obviously not used here in sg-core).

build/Dockerfile
leifmadsen: Yea I've been trying to adjust these types of emails as I find them.

leifmadsen:
I was originally going to move this to CentOS8 since I knew the downstream was using UBI8 (RHEL8 based), but then after some searching I realized I could use UBI directly due to a lack of dependency requirements, and found the unauthenticated access to the container image allowing builds to happen.

This will better align this project for import into downstream.

leifmadsen:
The testing here I'm adding is going to fail (it fails locally). We will likely have to address this in another commit, but for now I wanted to get things scaffolded out.

leifmadsen:
[Migrate sg-core to ubi8](https://trello.com/c/6qq4JdjV/1179-migrate-sg-core-to-ubi8)

PR #16, opened 2020-11-16
Matching golang version in test script

csibbitt:
Test script was running `go mod vendor` with golang 1.13 from the default Travis environment, with the source dir mounted as a volume. This removes the "explicit" tags in modules.txt [1], causing the build to fail with "	collectd.org@v0.3.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt" when it is later built for real with golang 1.14 [2]. This script should probably run tests using the same version of golang we will eventually build with.  I'm removing this line because we do the `go mod vendor` before committing now. 

[1] https://github.com/infrawatch/sg-core/blob/master/vendor/modules.txt#L2
[2] https://travis-ci.org/github/infrawatch/sg-core/builds/743489014#L1237

### DIFF
--- a/build/test-framework/run_tests.sh
+++ b/build/test-framework/run_tests.sh
@@ -16,6 +16,14 @@ go get -u github.com/mattn/goveralls
 go get -u golang.org/x/lint/golint
 go get -u honnef.co/go/tools/cmd/staticcheck
 
+# Update to golang which matches version in ubi:8
+go get golang.org/dl/go1.14.7
+go1.14.7 download
+alias go=go1.14.7
+
+# Add test deps to vendor dir
+go mod vendor
+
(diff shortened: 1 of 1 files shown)


csibbitt:
Problem shifted to mismatches with the testing requirements, so pursuing the idea of actually using the correct golang version instead.

pleimer (approved):
It will be fun keeping these versions in sync

leifmadsen:
> It will be fun keeping these versions in sync

There is probably something we could do where we just use a container to build the artifacts for us or something, then we could just re-leverage the same base image (and thus, keep our Golang deps in sync automatically).

leifmadsen (approved):

PR #19, opened 2020-11-26
Initial round of passing lint tests in sg-core

leifmadsen:
Basing this on top of the current github-actions branch. This should be rebased to master once / if that merges (or changes).

### DIFF
--- a/pkg/collectd/collectd.go
+++ b/pkg/collectd/collectd.go
@@ -7,7 +7,7 @@ import (
 	jsoniter "github.com/json-iterator/go"
 )
 
-type CollectdMeta struct {
+type collectdMeta struct {
 	X map[string]interface{} `json:"-"` // Rest of the fields should go here.
 }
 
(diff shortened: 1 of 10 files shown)


leifmadsen:
[Migrate from Travis-CI to GitHub Actions](https://trello.com/c/QVFdw42L/1210-migrate-from-travis-ci-to-github-actions)

pkg/collectd/collectd.go
```diff
@@ -7,7 +7,7 @@ import (
 	jsoniter "github.com/json-iterator/go"
 )
 
-type CollectdMeta struct {
+type collectdMeta struct {
```
pleimer, line 10: 👍

The performance of this should probably be analysed more in depth. Golang object allocation can be expensive time-wise. This is not for this PR but I will some testing for the SG core refactor

pleimer (approved):
LGTM

pkg/collectd/collectd.go
leifmadsen: I wonder if there is some profile tooling we could use to help us analyse these types of code paths?

...there is... I did some digging, and here are a few things we can look in to as part of the refactoring:

* https://goreportcard.com/report/github.com/infrawatch/sg-core
* https://blog.golang.org/pprof
* https://github.com/golangci/golangci-lint

PR #20, opened 2020-12-02
Add golangci-lint action to CI

leifmadsen:


pkg/cacheutil/cacheserver_test.go
```diff
@@ -70,7 +70,10 @@ func TestCacheExpiry(t *testing.T) {
 	cs := NewCacheServer()
 	ctx := context.Background()
 
-	go cs.Run(ctx)
+	// TODO: we should be capturing the error not ignoring it
+	go func() {
+		_ = cs.Run(ctx)
```
pleimer, line 75: you can just make it:
```
go func() {
    err := cs.Run(ctx)
    assert.Ok(t, err)
}
```

pkg/unixserver/server.go
```diff
@@ -420,7 +420,11 @@ func Listen(ctx context.Context, address string, w *bufio.Writer, registry *prom
 
 	// cache server
 	cache := cacheutil.NewCacheServer()
-	go cache.Run(ctx)
+
+	// TODO: check for error returns properly
+	go func() {
+		_ = cache.Run(ctx)
```
pleimer, line 426: same as above

pkg/unixserver/server_test.go
```diff
@@ -29,7 +29,10 @@ func TestCDMetrics(t *testing.T) {
 		cs.Interval = 1
 		ctx := context.Background()
 
-		go cs.Run(ctx)
+		// TODO: check for error returns properly
+		go func() {
+			_ = cs.Run(ctx)
```
pleimer, line 34: same as above

pleimer (approved):

PR #25, opened 2021-02-24
CI fixes

pleimer:


plugins/handler/ceilometer-metrics/main.go
```diff
@@ -241,7 +241,7 @@ func (c *ceilometerMetricHandler) Config(blob []byte) error {
 	return nil
 }
 
-//New ceilometer metric handler constructor
+//  New ceilometer metric handler constructor
```
paramite, line 244: two ws

pleimer:
There's still a couple linter issues that need to be resolved here and I have to figure out the problem with the Linter running in GHA

pleimer:
I have to implement a manual install of the qpid-proton development files in GHA because the ubuntu packages did not install the files where expected. This results in a really long runtime, unfortunately, and most likely a much higher GHA quota usage. Hopefully there is a way to work around this. I will research https://github.com/actions/cache for potential caching

pleimer:
I think this container based approach may actually be the best option: https://www.edwardthomson.com/blog/github_actions_5_building_in_containers.html

PR #28, opened 2021-03-03
Add logs handler and Loki application plugins

vyzigold:
This pull request adds logs and Loki support to sg-core. Most of the code (commits from "Add ending 0..." to "Update the loki...") you were already able to see at [[1]](https://github.com/pleimer/sg-core/pull/18). Commit "Cleanup" addresses the issues mentioned in the comments for that PR. The last 2 commits mostly just modify the code to use the current event bus.

I borrowed and modified the function for parsing time from [plugins/handler/events/pkg/lib/time.go](https://github.com/infrawatch/sg-core/blob/master/plugins/handler/events/pkg/lib/time.go). I think we might want to move this one to apputils/system, I think it could be useful in many more situations.

I also have a question about severity. As you can see from the code, syslog distinguishes between more severity values, than our bus currently allows. I tried to somehow map the syslog severity to our severity, which might be good enough. But maybe we should consider adding some of the syslog severity values to the severity on the event bus.

[[1]](https://github.com/pleimer/sg-core/pull/18) [https://github.com/pleimer/sg-core/pull/18](https://github.com/pleimer/sg-core/pull/18)


pleimer (commented):
Looks great! I am not at all opposed to moving some of the utility functions to the apputils library. We could also consider adding more severity statuses, but in this patch, your mappings make total sense to me. Ceilometer events were similar in that there are a ton of severities that really only boil down to those 3 categories

Couple comments, but I think this almost ready to merge

plugins/handler/logs/pkg/lib/time.go
```diff
@@ -0,0 +1,22 @@
...
+	"fmt"
+)
+
+var (
+	isoTimeLayout = "2006-01-02 15:04:05.000000"
+	rFC3339       = "2006-01-02T15:04:05.000000"
+)
+
+//EpochFromFormat get epoch time from one of select time string formats
+func TimeFromFormat(ts string) (time.Time, error) {
```
pleimer, line 14: Linter error

plugins/handler/logs/main.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// remove message and timestamp from labels (leave the rest)
+	delete(logFields, l.config.MessageField)
+	delete(logFields, l.config.TimestampField)
+
+	parsedLog = data.Event {
+		Index: index,
+		Time: timestamp,
+		Type: data.LOG,
+		Publisher: "sg-core",
```
pleimer, line 108: Does the rsyslog message contain information about what node the message came from? That would be more useful than 'sg-core' here

plugins/transport/socket/main.go
```diff
@@ -81,6 +81,7 @@ func (s *Socket) Run(ctx context.Context, w transport.WriteFn, done chan bool) {
 	go func() {
 		for {
 			n, err := pc.Read(msgBuffer)
+			msgBuffer[n] = 0
```
pleimer, line 84: You found a pretty critical bug here that exists in master right now (woops). Could you replace the call to the write function with `w(mspBuffer[:n])` instead of setting a null value? Just so the write function stack size can be minimized when the buffer isn't full

pleimer: Props on catching that, I don't know how that made it past all the testing I did, but I guess that is why unit tests are so important (plus a working CI pipeline)

plugins/application/loki/main.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// ReceiveEvent ...
+func (l *Loki) ReceiveEvent(log data.Event) {
+	switch log.Type {
+	case data.LOG:
+		lokiLog, err := lib.CreateLokiLog(log)
+		if err != nil {
+			l.logger.Metadata(logging.Metadata{"plugin": "loki", "log": log, "error": err})
+			l.logger.Info("failed to parse the data in event bus - disregarding")
```
paramite, line 47: Maybe this could be warning instead. If the parsing failure is also info in other apps, then ignore this :).

plugins/application/loki/main.go
```diff
@@ -0,0 +1,90 @@
...
+			return
+		}
+		l.logChannel <- lokiLog
+	default:
+		l.logger.Metadata(logging.Metadata{"plugin": "loki", "event": log})
+		l.logger.Info("received event data (instead of log data) in event bus - disregarding")
```
paramite, line 53: Ditto.

plugins/application/loki/main.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+//Run run loki application plugin
+func (l *Loki) Run(ctx context.Context, done chan bool) {
+	l.logger.Metadata(logging.Metadata{"plugin": "loki", "url": l.config.Connection})
+	l.logger.Info("storing logs to loki.")
```
paramite, line 60: Definitely Debug. We don't want to pollute log with such messages ... there will be plenty of those :).

plugins/application/loki/pkg/lib/loki.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/infrawatch/sg-core/pkg/data"
+)
+
+type LokiConfig struct {
+	Connection  string `validate:"required"`
+	BatchSize   int
+	MaxWaitTime int
+}
+
+// Creates labels used by Loki. Discards everything, that isn't
```
pleimer, line 17: Would this not leave out any important information? I'm not as familiar with the structure of incoming logs, but I wonder if it would be better to just convert everything to a string type

plugins/application/loki/pkg/lib/loki.go
```diff
@@ -0,0 +1,51 @@
...
+// a string
+func createLabels(rawLabels map[string]interface{}) (map[string]string, error) {
+	result := make(map[string]string)
+	for key, value := range rawLabels {
+		if typedValue, ok := value.(string); ok {
+			result[key] = typedValue
+		}
+	}
+	if len(result) == 0 {
+		return nil, fmt.Errorf("Unable to create log labels")
```
pleimer, line 27: Error message shouldn't be capatalized

plugins/application/loki/pkg/lib/loki.go
```diff
@@ -0,0 +1,51 @@
...
+
+func CreateLokiLog(log data.Event) (connector.LokiLog, error) {
+	labels, err := createLabels(log.Labels)
+	if err != nil {
+		return connector.LokiLog{}, err
+	}
+
+	msg, ok := log.Annotations["log_message"].(string)
+	if !ok {
+		return connector.LokiLog{}, fmt.Errorf("Unable to locate the log message")
```
pleimer, line 40: ^^

plugins/handler/logs/main.go
paramite: Agree, use sg-core only when hostname is not there.

plugins/handler/logs/pkg/lib/time.go
paramite: s#//EpochFromFormat#//TimeFromFormat#

plugins/application/loki/pkg/lib/loki.go
```diff
@@ -0,0 +1,51 @@
...
+	return result, nil
+}
+
+func CreateLokiLog(log data.Event) (connector.LokiLog, error) {
+	labels, err := createLabels(log.Labels)
+	if err != nil {
+		return connector.LokiLog{}, err
+	}
+
+	msg, ok := log.Annotations["log_message"].(string)
```
pleimer, line 38: The only problem with the Loki plugin expecting the "log_message" index in annotations is that handlers have to create messages specifically for storage into loki. I think it's fine for now, but in the future it may be worth either adding a 'message' field to events or creating an event subtype for logs that contains the field

plugins/transport/socket/main.go
paramite: Kudos indeed. Just out of curiosity (since I have no idea what's wrong here), what the last byte contains?

paramite (changes requested):

plugins/handler/logs/main.go
pleimer: That would be your hostname variable above most likely

plugins/application/loki/pkg/lib/loki.go
paramite: Actually, adding Message attribute to the Event is good idea ... we could then send actual tags for error events in labels and use message instead ... maybe even completely skip adding "description" to annotations and send it empty.

plugins/handler/logs/main.go
leifmadsen: And make sure you're sending the full hostname for multi-cloud compatibility. We already have that fix required for libpodstats :D

plugins/transport/socket/main.go
pleimer: The message buffer isn't cleared after a message is consumed, so if a new, shorter message arrives, there can be remnants of the old one left after the new message ends. Since the socket plugin was writing the whole buffer to the handler every time, the handler would get corrupted data

leifmadsen: Also assume this is `w(msgBuffer[:n])` without the typo heh

plugins/handler/logs/main.go
pleimer: 😀 calling me out, I see. In this case, it's up to rsyslog sending the fqdn though

plugins/transport/socket/main.go
pleimer: If only comments didn't tell everyone that you edited it

plugins/handler/logs/main.go
leifmadsen: I didn't post blame for anyone. You took that all on your own with this comment lol

pleimer:
I needed to make a couple of bug fixes in #29 and I included the socket buffer change there, so it you can just merge that in here when you are ready

plugins/handler/logs/main.go
vyzigold: I wasn't sure what exactly Publisher is. The hostname variable should work as long as rsyslog is configured correctly.

plugins/application/loki/pkg/lib/loki.go
vyzigold: The current log format sent by our rsyslog configuration can be seen [here](https://github.com/infrawatch/sg-core/blob/8b6e7586c7c396cf81d57972a3d346d18296c79c/plugins/transport/dummy-logs/main.go#L19). My thinking was, that everything we receive is a string, so this in fact shouldn't discard anything. I actually think, we might have too many labels.

plugins/application/loki/main.go
vyzigold: I looked at other app plugins, and the only similar line to this I found was in [elastic search](https://github.com/infrawatch/sg-core/blob/90757690b2ab6a4495db8ad8e47d6c35eff99397/plugins/application/elasticsearch/main.go#L68). And it's an error there, so which one is better?

plugins/application/loki/pkg/lib/loki.go
vyzigold: Didn't know this.

vyzigold: I actually found a few more capitalized error messages

plugins/application/loki/pkg/lib/loki.go
vyzigold: From logging point of view a Message attribute in Event would be really good.

paramite: If you think you can handle this change, feel free to add it to this PR ... since it is logging-connected change anyway.

plugins/application/loki/main.go
paramite: If ES has error, let's keep consistency and have it as error here too.

plugins/application/loki/pkg/lib/loki.go
pleimer: If in the future, a different log reporter uses an Int value for one of the labels or if one of the handlers adds a label that has a a value of something other than a string, it may lead to confusion as to why they aren't showing up. You could just use a similar function to Martin's assimilateMap func: https://github.com/infrawatch/sg-core/blob/8b6e7586c7c396cf81d57972a3d346d18296c79c/plugins/application/alertmanager/pkg/lib/utils.go#L16

pleimer: If we wanted to omit some of the labels, that would be the responsibility of the log handler to filter some out

vyzigold: I added the conversion with Martin's assimilateMap function.

plugins/application/loki/pkg/lib/loki.go
vyzigold: I added the Message attribute to the Event and used it in the logs handler and Loki application.

plugins/application/loki/main.go
vyzigold: I changed both to Error

vyzigold:
I think I made all the required changes, so unless there is anything else, we could merge this?

vyzigold:
the assimilateMap is probably also something, that could be moved to apputils/system

paramite:
Definitely agree, maybe not apputils/system, but something like apputils/misc, but feel free to submit additional duo of PRs for apputils and sg-core, making this happen ;).

PR #33, opened 2021-03-05
Clean Dockerfile

pleimer:


leifmadsen:
Looks like something is wrong in the `go.sum` causing the linting to fail and such. Do we expect all the tests to be passing here?

Can we add a test job that checks for builds? (unless I just missed it).

pleimer:
@leifmadsen no I am fixing all of the tests in  #25 so they are not expected to pass here. Linting is failing for a number of reasons that I am fixing there

leifmadsen:
> @leifmadsen no I am fixing all of the tests in #25 so they are not expected to pass here. Linting is failing for a number of reasons that I am fixing there

ship it :)

PR #34, opened 2021-03-05
Elasticsearch log storage

pleimer:
store logs in the same manner as events in ES. This PR is blocked by #28


paramite (changes requested):
Logs are not events, so this seems to me like using cannon to kill a sparrow. I understand that from the point of implementation side it's good enough, but in case of data processing it's whole inner map just for single key. I believe we want to make not only code, but also data cleaner with the refactor. So maybe a byt preprocessing in data.LOG branch and then let it drop to data.EVENT?

pleimer:
Are you saying there will be maps inside of maps? I am not 100% sure what you mean, although I do expect that this PR will need to change when #28 merges. Feel free to propose or push up to this branch too, I just wanted to open this up to get the work for ES started since this is what we need for 1.3

vyzigold:
I don't know, how exactly ES works in comparison with Loki, but at minimum, will need to take into account the change I made to the Event in #28. (Take the log message from Event.Message and save it to ES too.)

paramite:
@pleimer That was a valid comment when the log message was in Annotations, eg. you would save following to ES:

{<snip>
  "labels": {"a": "b", "c":"d"},
  "annotations": {"log_message": "error! scream!"}
}

while it would be cleaner to save following:
{<snip>
  "labels": {"a": "b", "c":"d"},
  "message": "error! scream!"
}

But as Jaromir mentioned, this is not a case, because we now have log record in Event.Message attribute, so we need to form above structure for ES. Eg. for logs we don't need annotations (there won't be any, just labels and timestamp and the message).

paramite:
@vyzigold ES eats whatever you tell it to eat :) ... but that does not matter we should not care what it eats :D

pleimer:
Untested at the moment. I want to test these changes with rsyslog before this is merged

paramite:
Whoops, tested log saving can be found at https://github.com/infrawatch/sg-core/pull/38

PR #39, opened 2021-03-18
No error when non critical fields are not present

pleimer:
https://github.com/infrawatch/smart-gateway/pull/100 made me realise the errors being thrown when ceilometer metrics did not contain some of the fields is not the behaviour we want. Instead of an error being generated, sg-core should just skip those fields and send the rest

### DIFF
--- a/plugins/handler/ceilometer-metrics/main.go
+++ b/plugins/handler/ceilometer-metrics/main.go
@@ -107,11 +107,7 @@ func (c *ceilometerMetricHandler) Handle(blob []byte, reportErrs bool, mpf bus.M
 		}
 
 		mType := ceilTypeToMetricType[m.CounterType] // zero value is UNTYPED
-
-		cNameShards := strings.Split(m.CounterName, ".")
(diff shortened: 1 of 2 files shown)


plugins/handler/ceilometer-metrics/main.go
```diff
@@ -206,31 +177,48 @@ func genLabels(m ceilometer.Metric, publisher string, cNameShards []string) ([]s
...
 
-	labelKeys[4] = "project"
-	labelVals[4] = m.ProjectID
+	labelKeys[2] = "type"
+	labelVals[2] = ctype
 
-	labelKeys[5] = "unit"
-	labelVals[5] = m.CounterUnit
+	// non - critical
+	index := 3
```
paramite, line 191: Hmm, I thought counter_* values are always present in ceilometer metric payload.

paramite (approved):

plugins/handler/ceilometer-metrics/main.go
pleimer: You may be right, I did not attempt to confirm it

Issue #44, opened 2021-03-30
Ceilometer-metrics plugin not providing correct labels for Prometheus


pleimer:
Valid complaint from Prometheus:
`2021-03-30 15:20:21 [ERROR] prometheus failed scrapping metric [plugin: Prometheus, error: "" is not a valid label name for metric "ceilometer_volume_size"]`

The (cleaned) incoming message from Ceilometer
```
{
  "request": {
    "oslo.message": {
      "priority": "SAMPLE",
      "event_type": "metering",
      "timestamp": "2021-03-30 15:20:19.900127",
      "publisher_id": "telemetry.publisher.controller-1.osp-cloudops-06",
      "payload": [
        {
          "counter_name": "volume.size",
          "resource_id": "f1cd28dc-74ec-4141-90de-06aaf7e621a4",
          "timestamp": "2021-03-30T15:20:19.891893",
          "counter_volume": 2,
          "user_id": "9c6d2e7419794be7b89964d886b839b8",
          "message_signature": "81f024f868412b7fd5558fa695279dc5a5e6237bb12f8c0be7d3bc60ec9f3c5a",
          "resource_metadata": {
            "status": "in-use",
            "migration_status": null,
            "attachments": [
              {
                "server_id": "e3126196-b2ad-4a92-95c0-9dcb65e86553",
                "attachment_id": "e4f5cc4f-fb57-4299-ab06-6a6600d7a143",
                "attached_at": "2021-03-27T14:47:48.000000",
                "host_name": null,
                "volume_id": "f1cd28dc-74ec-4141-90de-06aaf7e621a4",
                "device": "/dev/vda",
                "id": "f1cd28dc-74ec-4141-90de-06aaf7e621a4"
              }
            ],
            "os-vol-host-attr:host": "hostgroup@tripleo_ceph#tripleo_ceph",
            "source_volid": null,
            "volume_type": "tripleo",
            "snapshot_id": null,
            "name": ""
          },
          "source": "openstack",
          "counter_unit": "GB",
          "project_id": "db3fce7b7aeb4109bb2794f9337e68fa",
          "message_id": "7071168c-916b-11eb-a587-525400f00aad",
          "monotonic_time": null,
          "counter_type": "gauge"
        },
        {
          "counter_name": "volume.size",
          "resource_id": "ed8102c3-923a-4f5a-9a24-d59afc174755",
          "timestamp": "2021-03-30T15:20:19.891893",
          "counter_volume": 2,
[...]

PR #45, opened 2021-03-30
Debug enhancements

pleimer:
- Add configuration option to socket plugin that enables it to dump the bytes it receives as-is to a file or console
- Fix https://github.com/infrawatch/sg-core/issues/44
- Make expiry multiple parameter configurable in the Prometheus plugin. This number sets the interval at which this plugin polls metrics to see if they have expired to a multiple of the interval specified in the metric.
- Move expiration notice to logging level `debug` now that debugging is enabled in the SGO

### DIFF
--- a/plugins/application/prometheus/main.go
+++ b/plugins/application/prometheus/main.go
@@ -25,9 +25,10 @@ var typeToPromType map[data.MetricType]prometheus.ValueType = map[data.MetricTyp
 }
 
 type configT struct {
-	Host          string
-	Port          int  `validate:"required"`
-	WithTimestamp bool `yaml:"withTimeStamp"`
+	Host               string
+	Port               int  `validate:"required"`
+	WithTimestamp      bool `yaml:"withTimeStamp"`
+	ExpirationMultiple int  `yaml:"expirationMultiple"` // multiple of metric interval at which that metric should be expired (seconds)
 }
 
 // used to expire stale metrics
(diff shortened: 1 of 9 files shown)


plugins/application/prometheus/main.go
```diff
@@ -168,6 +174,7 @@ func (pc *PromCollector) UpdateMetrics(name string, time float64, typ data.Metri
 	cacheKey := pc.cacheindexbuilder.String()
 	mProcItf, found := pc.mProc.Load(cacheKey)
 	if !found {
+		// pc.logger.Infof("creating new metric with cache key '%s'", cacheKey)
```
leifmadsen, line 177: Is this something you no longer need, or a comment showing how to use this? If this is code that is no longer being used, please remove vs commenting out.

plugins/handler/ceilometer-metrics/main_test.go
```diff
@@ -47,6 +48,18 @@ func MetricReceive(name string, mTime float64, mType data.MetricType, interval t
...
+// 	CounterName      string  `json:"counter_name"`
+// 	CounterType      string  `json:"counter_type"`
+// 	CounterUnit      string  `json:"counter_unit"`
+// 	CounterVolume    float64 `json:"counter_volume"`
+// 	UserID           string  `json:"user_id"`
+// 	ProjectID        string  `json:"project_id"`
+// 	ResourceID       string  `json:"resource_id"`
+// 	Timestamp        string
+// 	ResourceMetadata metadata `json:"resource_metadata"`
+// }
```
leifmadsen, line 62: No longer need this struct?

pleimer: that was for reference my mistake

plugins/application/prometheus/main.go
pleimer: deleting it

PR #47, opened 2021-05-06
[logging] Integration tests

paramite:
This patch sets structure for integration tests starting with logging
plugins of sg-core

Changelog:
 - enables unit tests for loki application
 - refactors unit tests execution to centos image to enable spawning
   loki container correctly
 - introduction of integration tests structure
 - itegration tests for logs handler and elasticsearch and loki applications


.github/workflows/tests.yml
```diff
@@ -23,32 +30,35 @@ jobs:
...
-        run: go test -v -coverprofile=profile.cov ./...
-      - name: Send coverage
-        uses: shogo82148/actions-goveralls@v1
-        with:
-          path-to-profile: profile.cov
+      - name: Checkout code
+        uses: actions/checkout@v2
+      # start services that needs additional configuration and builtin services
+      # list does not work in those cases
+      - name: Start Loki service
```
pleimer, line 49: Curious why do you run the Loki service here?

pleimer (approved):
LGTM once the issue with the unit tests is resolved

.github/workflows/integration.yml
```diff
@@ -0,0 +1,73 @@
...
+  TEST_IMAGE: centos:8
+on: push
+
+jobs:
+  logging-integration-tests:
+    name: [logging] handler: logs; application: elasticsearch, loki
+    runs-on: ubuntu-20.04
+    services:
+      elastic:
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.5.1
```
leifmadsen, line 22: We can bump this to 7.10.2 now that we're using that in STF

.github/workflows/integration.yml
```diff
@@ -0,0 +1,73 @@
...
+
+  QDR_IMAGE: quay.io/interconnectedcloud/qdrouterd:1.15.0
+  QDR_VOLUME: "--volume=${{ github.workspace }}/ci/service_configs/qdr:/etc/qpid-dispatch:ro"
+  QDR_PORT: "-p 5666:5666"
+
+  LOKI_IMAGE: grafana/loki:2.1.0
+  LOKI_VOLUME: "--volume=${{ github.workspace }}/ci/service_configs/loki:/etc/loki:ro"
+  LOKI_PORT: "-p 3100:3100"
+
+  TEST_IMAGE: centos:8
```
leifmadsen, line 13: Can this be moved to ubi8:latest instead? Otherwise we should consider moving to centos stream

.github/workflows/tests.yml
```diff
@@ -1,6 +1,13 @@
 name: CI
 env:
-  COVERALLS_TOKEN: ${{ secrets.COVERALLS_TOKEN }}
+  PROJECT_ROOT: /root/go/src/github.com/infrawatch/sg-core
+
+  LOKI_IMAGE: grafana/loki:2.1.0
```
leifmadsen, line 5: For version tags, might be ideal to put these all in one spot and refer to them via env vars just to make it easier to update this in the future.

ci/integration/logging/run_sg.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/env bash
+# CI script for CentOS8 job
+# purpose: spawn sg-core to process messages sent by rsyslog
+
+set -ex
+
+# enable required repo(s)
+yum install -y epel-release
```
leifmadsen, line 8: It would be better if we could avoid using epel-release and base this on ubi8 which is going to better align everything downstream for us. I recently migrated a couple of the other repositories to ubi8 and there should be some yum repositories loaded that will help cover things that don't exist in native ubi8.

ci/integration/logging/run_sg.sh
```diff
@@ -0,0 +1,38 @@
...
+yum install -y git golang gcc make glibc-langpack-en qpid-proton-c-devel
+
+export GOBIN=$GOPATH/bin
+export PATH=$PATH:$GOBIN
+
+SOCKET=/tmp/rsyslog-test-socket
+BRIDGE_LOG=/var/log/sg-bridge.log
+
+# install and start sg-bridge
+git clone https://github.com/infrawatch/sg-bridge.git
```
leifmadsen, line 20: Cloning and building this each time seems a bit heavy no? Is there a dependency here where you need to build sg-bridge? If there is a cross-project interface here, I would like to see it try and clone the same-named branch like we do for STF CI stuff, with a fallback to `master` branch.

ci/integration/logging/sg_config.yaml
```diff
@@ -0,0 +1,30 @@
...
+      hostURL: http://127.0.0.1:9200
+      useTLS: false
+      useBasicAuth: false
+      bufferSize: 37
+      bulkIndex: true
+  - name: loki
+    config:
+        connection: http://127.0.0.1:3100
+        batchsize: 3
+        maxwaittime: 1
```
leifmadsen, line 30:
```suggestion
      connection: http://127.0.0.1:3100
      batchsize: 3
      maxwaittime: 1
```

vyzigold:
It looks like adding data.DEBUG and translating severity from number to string broke most of the log/event unit tests, but otherwise the PR looks ok to me.

paramite:
Yes indeed. This fact I missed and working on fixing it :)

paramite:
Ok unit tests are fine now, but coveralls report seems missing token. Investigating.

paramite:
I wasn't able to move the coverage reporting to unit test script and then realize that I'm actually trying to reinvent the wheel. It's fixed now as you can check at [1]. Unit tests fail because on PR from fork coveralls token is not passed. Going to work on comments now. 

[1] https://github.com/infrawatch/sg-core/actions

paramite:
If I did not missed anything I think this PR is good for next round of review and eventual merge.

.github/workflows/integration.yml
leifmadsen: I'd have preferred ubi8 here since that's what we're going to be building against downstream, but since I called out stream I guess this is ok.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,75 @@
...
+
+on: push
+
+jobs:
+  logging:
+    name: "[logging] handler: logs; application: elasticsearch, loki"
+    runs-on: ubuntu-20.04
+    services:
+      elastic:
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.10.2
```
leifmadsen, line 23: Move this in to an ENV like other images.

.github/workflows/tests.yml
leifmadsen: Disregard this one. Looks fine.

.github/workflows/tests.yml
```diff
@@ -23,32 +31,37 @@ jobs:
...
           version: v1.33
-  test-framework:
+  unit-tests:
     name: Unit tests
     runs-on: ubuntu-20.04
-    container: 'quay.io/plmr/sg-core-ci'
     services:
       elastic:
-        image: docker.elastic.co/elasticsearch/elasticsearch:7.5.1
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.10.2
```
leifmadsen, line 39: Move this to an ENV like other images.

ci/integration/logging/run_sg.sh
leifmadsen: Yea check this out in smart-gateway repo where I use opstools.repo for qpid-proton-c-devel.

https://github.com/infrawatch/smart-gateway/blob/master/build/test-framework/run_tests.sh

All the other repos we have are on ubi8, and our base image downstream is ubi8, so I'm kind of wanting it all to be in sync as much as possible so we don't get surprised in testing or builds when importing downstream.

ci/integration/logging/run_sg.sh
leifmadsen: I think the simplest thing is to basically check if the same-named branch exists on the sg-bridge side (for coordinated testing if changes on sg-bridge are required to satisfy a change here), and if not, then just pull from quay.io/infrawatch/sg-bridge:latest

It's not that big of a deal, but it could certainly be a bit faster test, although I suspect sg-bridge doesn't ultimately take all that long to deploy.

If you want to make the changes that's fine, if not, I won't block it.

ci/integration/logging/run_rsyslog.sh
```diff
@@ -0,0 +1,35 @@
...
+name=Fedora - ELN BaseOS - Developmental packages for the next Enterprise Linux release
+baseurl=https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose/BaseOS/\$basearch/os/
+#metalink=https://mirrors.fedoraproject.org/metalink?repo=eln&arch=\$basearch
+enabled=1
+gpgcheck=0
+skip_if_unavailable=False
+EOF
+
+# Locale setting in CentOS8 is broken
+yum install -y glibc-langpack-en rsyslog-omamqp1
```
leifmadsen, line 21: I think both ubi8, centos:streams(?) and Fedora all use `dnf` now actually.

leifmadsen (approved):
Could small comments, but otherwise this looks good to me. Thanks!

.github/workflows/integration.yml
paramite: Ah I see, ok will do.

.github/workflows/tests.yml
paramite: Will do.

ci/integration/logging/run_sg.sh
paramite: Oh I was not aware that we have this dependency in opstools repo. In that case I can move to ubi8. Thanks for a pointer.

ci/integration/logging/run_sg.sh
paramite: So ... you want me to run sg-bridge in separate container instead of sg-core container and in case the sg-bridge repo does not have the same topic branch use the prebuilt image ... got it.

ci/integration/logging/run_rsyslog.sh
paramite: Old habits die hard :)

paramite:
@leifmadsen I hope that you will be absolutely happy now :)

ci/integration/logging/run_sg.sh
```diff
@@ -0,0 +1,27 @@
...
+module_hotfixes=1
+EOF
+
+dnf install -y git golang gcc make qpid-proton-c-devel
+
+export GOBIN=$GOPATH/bin
+export PATH=$PATH:$GOBIN
+
+# install sg-core and start sg-core
+go mod tidy
```
leifmadsen, line 23: Do we need a test at some point that if `go mod tidy` actually results in changes, that we should fail on that?

ci/integration/logging/run_validation.sh
```diff
@@ -0,0 +1,107 @@
...
+HOST=$(hostname)
+
+ELASTIC_URL=http://127.0.0.1:9200
+LOKI_URL=http://127.0.0.1:3100
+
+######################### validate elasticsearch data #########################
+# debug output of cluster status
+curl -sX GET "$ELASTIC_URL/_cluster/health?pretty"
+# verify expected index
+#TODO(mmagr): adapt elasticseatch plugin to create index templates avoinding unnecessary prefix and suffix
```
leifmadsen, line 20:
```suggestion
#TODO(mmagr): adapt elasticsearch plugin to create index templates avoiding unnecessary prefix and suffix
```

ci/integration/logging/run_validation.sh
```diff
@@ -0,0 +1,107 @@
...
+# debug output of index content
+curl -sX GET "$ELASTIC_URL/${expected_index}/_search?pretty" -H 'Content-Type: application/json' -d'
+{
+  "query": {
+    "match_all": {}
+  }
+}
+'
+# verify expected documents
+res=$(curl -sX GET "$ELASTIC_URL/${expected_index}/_search" -H 'Content-Type: application/json' -d"
```
leifmadsen, line 34: Would it just make more sense to use single-quotes for the query so you don't have to escape the double-quotes? Maybe there is something going on here I don't understand though.

ci/unit/run_tests.sh
```diff
@@ -0,0 +1,24 @@
...
+[centos-opstools]
+name=opstools
+baseurl=http://mirror.centos.org/centos/8/opstools/\$basearch/collectd-5/
+gpgcheck=0
+enabled=1
+module_hotfixes=1
+EOF
+
+# without glibc-langpack-en locale setting in CentOS8 is broken without this package
+yum install -y git golang gcc make glibc-langpack-en qpid-proton-c-devel
```
leifmadsen, line 18:
```suggestion
dnf install -y git golang gcc make glibc-langpack-en qpid-proton-c-devel
```

leifmadsen (approved):

leifmadsen:
> @leifmadsen I hope that you will be absolutely happy now :)

Very close lol

ci/integration/logging/run_sg.sh
leifmadsen: PS not for fixing now, just asking. If you think so, we can just open an issue to add it as an enhancement.

paramite: Maybe. The reason of tidy is that for some reason without it build script was failing on missing dependencies, because the legacy go path mode was being used and the dependencies were not installed in go path. SO we definitely need some improvements on that side.

ci/integration/logging/run_validation.sh
paramite: That is a left over when I was trying to query log records not only by phrase, but also by host (saved in env variable) and by beginning of timestamp (also saved in env var). But I failed miserably on that front. Since I had al the queries written with double quotes double quotes. But yeah I'll fix it.

ci/integration/logging/run_sg.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/env bash
+# CI script for UBI8 job
+# purpose: spawn sg-core to process messages sent by rsyslog
+
+set -ex
+
+# enable required repo(s)
+cat > /etc/yum.repos.d/fedora-eln.repo <<EOF
```
leifmadsen, line 8: Huh I learned a new trick today that if you surround the EOF in single quotes like `'EOF'` it won't expand the `$` etc.

```
cat > /tmp/foo <<'EOF'
> $(bar) foo
> EOF

cat /tmp/foo 
$(bar) foo
```

leifmadsen:
I saw one of the tests failed, so I re-ran it just in case. I don't want to keep nit-picking at this. Looks good. As soon as all the tests pass feel free to merge this. Thanks!

paramite:
> I saw one of the tests failed, so I re-ran it just in case. I don't want to keep nit-picking at this. Looks good. As soon as all the tests pass feel free to merge this. Thanks!

Yes one of the metrics unit test is unstable and sometimes fail. We will have to figure this out in another patch.

PR #53, opened 2021-06-14
Fix concurrent conflicts in elasticsearch

paramite:
This patch makes use of concurrent package to fix conflicting writes/reads
to/from ES buffer when bulk index is configured

### DIFF
--- a/plugins/application/elasticsearch/main.go
+++ b/plugins/application/elasticsearch/main.go
@@ -13,6 +13,7 @@ import (
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
 
+	"github.com/infrawatch/sg-core/pkg/concurrent"
 	"github.com/infrawatch/sg-core/plugins/application/elasticsearch/pkg/lib"
 )
 
@@ -51,7 +52,7 @@ type Elasticsearch struct {
 	configuration *lib.AppConfig
 	logger        *logging.Logger
 	client        *lib.Client
-	buffer        map[string][]string
+	buffer        *concurrent.Map
 	dump          chan esIndex
 }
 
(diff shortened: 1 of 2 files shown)


csibbitt:
I wrote a similar patch this morning. https://github.com/infrawatch/sg-core/compare/csibbitt-205-buffer-concurrency-fix and was in the process of testing it when I saw this.

Can you explain why you chose concurrent.Map instead of a Mutex? Should I benchmark both?

paramite:
Hey, yes sure. Concurent package is using RWMutex too. It is just a piece of code we have in sg-core already [1], so I just used what we have.

[1] https://github.com/infrawatch/sg-core/blob/master/pkg/concurrent/concurrent.go#L7

paramite:
I'm pretty fine with your path though.

pleimer:
The other interesting one would to try would be sync.Map. However, it is only more performant in very specific circumstances (few cache misses, many CPUs) and I don't believe this situation is optimal for it.

csibbitt:
> The other interesting one to try would be sync.Map. However, it is only more performant in very specific circumstances (few cache misses, many CPUs) and I don't believe this situation is one of them

Honestly it wasn't until I read this that I realized this wasn't using sync.Map. I saw it quickly earlier and assumed that's what was going on (I agree that our use-case doesn't fit, I specifically opted not to use it). On re-read I see we've already solved this in a uniform way and I'd rather re-use that code if there is no significant performance penalty. So basically, I'm preferring this patch over https://github.com/infrawatch/sg-core/pull/54  if possible.

Thanks @paramite

csibbitt:
I tested this and it behaves the same as my patch. The performance differences should be minimal compared to other improvements we are contemplating, so I'll merge this version.

PR #55, opened 2021-06-22
Use the current apputils and update loki plugin

vyzigold:
This PR makes sg-core use the most recent version of apputils and does all the required changes to the Loki plugin, which is the only one affected by the changes in apputils. Thanks to this we should be able to run the Loki plugin inside the sg-core containers. (Currently it fails, because it unsuccessfully tries to find libproton.so, even though it actually doesn't need it).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/go-openapi/errors v0.20.0
 	github.com/go-playground/universal-translator v0.17.0 // indirect
 	github.com/google/uuid v1.2.0
-	github.com/infrawatch/apputils v0.0.0-20210325130739-482adc424cd9
+	github.com/infrawatch/apputils v0.0.0-20210621123139-68f3c7c25e4f
 	github.com/json-iterator/go v1.1.10
 	github.com/leodido/go-urn v1.2.1 // indirect
 	github.com/pkg/errors v0.9.1
--- a/plugins/application/loki/main.go
+++ b/plugins/application/loki/main.go
@@ -5,7 +5,7 @@ import (
 	"context"
 	"time"
 
-	"github.com/infrawatch/apputils/connector"
+	"github.com/infrawatch/apputils/connector/loki"
 	"github.com/infrawatch/apputils/logging"
 	"github.com/infrawatch/sg-core/pkg/application"
 	"github.com/infrawatch/sg-core/pkg/config"
@@ -25,7 +25,7 @@ type LokiConfig struct {
 // Loki plugin for forwarding logs to loki
 type Loki struct {
 	config     *LokiConfig
-	client     *connector.LokiConnector
+	client     *loki.LokiConnector
 	logger     *logging.Logger
 	logChannel chan interface{}
 }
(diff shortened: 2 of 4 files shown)


vyzigold:
The CI is currently failing because of a wrong API endpoint for checking Loki readiness, a fix for this is coming [here](https://github.com/infrawatch/sg-core/commit/a97f661e0240dc6d179382872c36f03b216283cf).

leifmadsen:
Is the expectation to cherry pick this back to stable-1.3 once it is complete?

vyzigold:
That depends if we ever want to use the Loki plugin with 1.3. Because right now in the upstream sg-core containers, when sg-core tries to load the Loki plugin, the whole sg-core crashes with: 
```
12:32:33 [ERROR] failed configuring application [application: loki, error: failed initializing application plugin: failed to open binary /usr/lib64/sg-core/loki.so: plugin.Open("/usr/lib64/sg-core/loki.so"): libqpid-proton-core.so.10: cannot open shared object file: No such file or directory]
```
(The Loki plugin doesn't even use that lib for anything.)
What is the timeline for 1.4?

leifmadsen:
> That depends if we ever want to use the Loki plugin with 1.3. Because right now in the upstream sg-core containers, when sg-core tries to load the Loki plugin, the whole sg-core crashes with:
> 
> ```
> 12:32:33 [ERROR] failed configuring application [application: loki, error: failed initializing application plugin: failed to open binary /usr/lib64/sg-core/loki.so: plugin.Open("/usr/lib64/sg-core/loki.so"): libqpid-proton-core.so.10: cannot open shared object file: No such file or directory]
> ```
> 
> (The Loki plugin doesn't even use that lib for anything.)
> What is the timeline for 1.4?

We can assume STF 1.4 won't show up during this calendar year. I expect to run with STF 1.3 for a while since standing up new y-stream releases requires so much time. I'm ok with slipping in new functionality to STF 1.3 at this point as long as documentation and testing are adequately covered.

leifmadsen:
I'm reconsidering my approval for backport, and perhaps working on getting STF 1.4 staged sooner rather than later, even if the functionality doesn't show up for several months. If we end up needing to backport this due to customer request, then we can deal with that at that point.

I'm hoping to get some automation setup in the next couple of weeks so that we can start making use of this functionality from upstream builds much more easily by creating a "stream" channel against the main branch. That work I'm hoping will start next week as I have a plan. That way we can start working these changes, and documenting them as we go, as finding all the features and doing a mass documentation lift at the end is really exhausting.

leifmadsen (approved):

vyzigold:
So do I understand it correctly, that we won't cherry-pick unless we are required to backport this? Should I merge this? :thinking:

leifmadsen:
> So do I understand it correctly, that we won't cherry-pick unless we are required to backport this? Should I merge this? thinking

Yea I think we'll merge this to main and then if we end up supporting any sort of logging on 1.3 we'll probably need to backport this at that time. Thanks!

Issue #12, opened 2020-07-16
[RFE] sg-core should be able to listen on more than one amqp address


leifmadsen:
This is a feature request.

sg-core should be able to listen on more than one amqp address at a time. It should also label the incoming metrics with the address, and any other metadata that would make it more clear which cloud the metrics are associated with.

paramite:
I believe that with the recent refactor this is true.

leifmadsen:
> I believe that with the recent refactor this is true.

I would think so yea. Is there a quick test that can get run at some point to verify this? Once verified feel free to close this out! Thanks!

vyzigold:
So, I've been thinking about this and decided to try it.

I used 2 different applications writing logs to 2 different amqp addresses. For each address I run a separate sg-bridge, each listening on one of the amqp addresses and sending logs to a different socket:
```
./bridge --amqp_url=amqp://127.0.0.1:5672/collectd --gw_unix=/tmp/smartgateway
```
```
./bridge --amqp_url=amqp://127.0.0.1:5672/logs --gw_unix=/tmp/logs
```

Then I configured sg-core to listen on both of the sockets at once (I used 2 socket transport plugins and 2 logs handlers) and after a quick small patch (each transport plugin requires a unique name) to sg-core I was successfully able to get logs from both amqp addresses to Loki. The required patch (along with my config) can be seen [here](https://github.com/vyzigold/sg-core/commit/a69c365fbac6a3b2f711c804754f86a6238e4e6e) (It will require a few more things if we want to use it).

The config I used:
```
pluginDir: /tmp/plugins
logLevel: debug
transports:
    - name: socket
      config:
          path: "/tmp/smartgateway"
      handlers:
      - name: logs
        config:
            timestampField: "@timestamp"
            messageField: "message"
            severityField: "severity"
            hostnameField: "host"
    - name: socket
      config:
          path: "/tmp/logs"
      handlers:
      - name: logs
        config:
            timestampField: "@timestamp"
            messageField: "message"
            severityField: "severity"
            hostnameField: "host"
applications: 
  - name: loki
    config:
        connection: "http://localhost:3100"
        maxwaittime: 5s
```
Do we want this functionality?

leifmadsen:
> Do we want this functionality?

Honestly I'm not sure lol. I think having the ability to do this would be great though! Obviously we'd have to verify that Smart Gateway Operator could also provide the ability to listen on multiple interfaces.

[...]

PR #59, opened 2021-06-29
Allow to listen on multiple sockets at once

vyzigold:
This PR implements unique names for transport plugins, which allows us to run multiple plugins of the same type (for example 2 socket transport plugins). It also adds some notes for this in the docs/ folder.

Solves #12.

### DIFF
--- a/cmd/manager/manager.go
+++ b/cmd/manager/manager.go
@@ -7,6 +7,8 @@ import (
 	"plugin"
 	"strings"
 	"sync"
+	"strconv"
+	"math/rand"
 
 	"github.com/infrawatch/apputils/logging"
 	"github.com/infrawatch/sg-core/pkg/application"
@@ -50,29 +52,30 @@ func SetLogger(l *logging.Logger) {
 }
 
 // InitTransport load tranpsort binary and initialize with config
-func InitTransport(name string, config interface{}) error {
+func InitTransport(name string, config interface{}) (string, error) {
 	n, err := initPlugin(name)
 	if err != nil {
-		return errors.Wrap(err, "failed initializing transport")
+		return "", errors.Wrap(err, "failed initializing transport")
 	}
 
 	new, ok := n.(func(*logging.Logger) transport.Transport)
 	if !ok {
-		return fmt.Errorf("plugin %s constructor 'New' did not return type 'transport.Transport'", name)
+		return "", fmt.Errorf("plugin %s constructor 'New' did not return type 'transport.Transport'", name)
 	}
 
-	transports[name] = new(logger)
+	nam := strconv.Itoa(rand.Intn(10000))
+	transports[nam] = new(logger)
 
 	c, err := yaml.Marshal(config)
 	if err != nil {
-		return errors.Wrapf(err, "failed parsing transport config for '%s'", name)
+		return "", errors.Wrapf(err, "failed parsing transport config for '%s'", name)
 	}
 
-	err = transports[name].Config(c)
+	err = transports[nam].Config(c)
 	if err != nil {
-		return err
+		return "", err
 	}
-	return nil
+	return nam, nil
 }
 
 // InitApplication initialize application plugin with configuration
--- a/cmd/manager/manager.go
+++ b/cmd/manager/manager.go
@@ -5,9 +5,9 @@ import (
 	"fmt"
 	"path/filepath"
 	"plugin"
+	"strconv"
 	"strings"
 	"sync"
-	"strconv"
 
 	"github.com/infrawatch/apputils/logging"
 	"github.com/infrawatch/sg-core/pkg/application"
--- a/cmd/manager/manager.go
+++ b/cmd/manager/manager.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"sync"
 	"strconv"
-	"math/rand"
 
 	"github.com/infrawatch/apputils/logging"
 	"github.com/infrawatch/sg-core/pkg/application"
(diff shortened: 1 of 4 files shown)


pleimer (approved):
Nice addition thanks for this

leifmadsen:
Very nice thanks for adding this. Are there any unit tests that could be added for this as well?

The main reason I haven't merged this is I want to make sure we track any other work that needs to support this, such as changes to STO/SGO so that we can make use of this, and figure out exactly how this could be used going forward.

I don't want to lose track of this functionality, so it would be good to flesh out how we can use it. Even if we don't do the integration right now, I'd like to make sure we figure out how we'd use it, and then file issues in the appropriate repositories so we can implement it.

leifmadsen:
Last thing, and this is really a procedure thing, is I'd like to figure out how we can track this type of functionality against a future STF release so that we can make sure the release notes cover this new feature. Might be good enough to create a single "release" BZ for downstream import tracking, and just keep bumping the Release Notes doc text. Will discuss with the team.

Hopefully we can figure that out shortly as I'd hate to see this code rot and not make it in. I do feel like adding unit tests and smoke tests would be a huge advantage though, so we might want to track that separately.

cmd/manager/manager.go
```diff
@@ -50,29 +51,32 @@ func SetLogger(l *logging.Logger) {
...
 	new, ok := n.(func(*logging.Logger) transport.Transport)
 	if !ok {
-		return fmt.Errorf("plugin %s constructor 'New' did not return type 'transport.Transport'", name)
+		return "", fmt.Errorf("plugin %s constructor 'New' did not return type 'transport.Transport'", name)
 	}
 
-	transports[name] = new(logger)
+	// Append the current length of transports
+	// to make each name unique
+	uniqueName := name + strconv.Itoa(len(transports))
```
csibbitt, line 67: :+1: I saw another commit where this was random, but this is better.

csibbitt (commented):
This code LGTM, but I'm hesitant to add any complexity without a clear indication why. I've read https://github.com/infrawatch/sg-core/issues/12 and I'm not sure I understand how this is intended to be used, or who would like to use it.

leifmadsen:
> This code LGTM, but I'm hesitant to add any complexity without a clear indication why. I've read #12 and I'm not sure I understand how this is intended to be used, or who would like to use it.

I think the idea here was to allow us to limit the number of SGs being deployed so that in a multi-cloud scenario we didn't need to instantiate 4 (now 5...) SGs per cloud. Probably as a precursor to being able to start thinking about horizontally scaling the SG instances based on received messages or something (which obviously requires much more work and coordination).

It was filed a year ago, so I think our approach / thinking may have changed a bit more to "that's kind of just how it is" wrt running many sg-core instances.

In our reference architecture of a 3-node OCP instance where the controllers are workload enabled, I can see wanting to support multiple small clouds with a single sg-core and many inputs. I'm just not sure if that's really a priority at the moment.

Not really sure what to do here ultimately. We could merge this as long as it doesn't seem like it would interfere with what we're currently doing, or we could say we won't support that functionality at all and abandon this? I don't really know what to suggest here.

leifmadsen:
Additional thought, is that we could possibly reduce our footprint a bit on the data collectors, where we could run a metrics sg-core (sg-bridge for ceilometer address, sg-bridge for collectd address) and then forward that to multiple handlers and a single application (Prometheus, ElasticSearch) for data storage, but I'm not sure this code is enough to achieve that either. While it's technically possible to load multiple handlers, I'm not sure if there is an ability to strongly correlate each of the sg-bridge transports to a specific handler.

Even if we could, it comes down to if that architecture is one we want to make use of or not.

CC: @pleimer

csibbitt:
I'm okay with merging this because I do like the prospect of one day running fewer SGs for deployments where that makes sense. I don't think we should be pointing it out in docs or anything at this point, though. I'm not concerned that we'll "lose track" of it; when we are ready for it the work is already complete. The remaining question is whether it's worth scoping all that additional work right now, and I think that without a strong use case or a limitation we're overcoming, it's not worth it yet. We might end up using this more in bench testing than anything else for the moment.

leifmadsen:
Yea I think I'm in agreement. I'm ok with merging this as well.

pleimer:
The original design already allows for multiple transports to run at the same time. This patch just enables more than one of the same transport, an oversight in the original implementation. There really isn't any added complexity here and this patch does not affect the way we currently use sg-core. 

@leifmadsen is right that there is no way to strongly correlate which messages come from which transport. There is no way of differentiating messages coming from socket1 and socket2 which means there could be interference if all of the labels happen to be the same (which is very likely). There's a few things we could do to address this

1. Have sg-bridge tack some bytes on the front or back of the message indicating which address a message is coming from. The handler plugins must be aware of this external protocol and extract the bytes before attempting to parse the JSON. The (pretty big) drawback to this is we lose the general usability of these handlers with any other transport type such as AMQP or HTTP since they now expect the bytes that sg-bridge adds to the message

2. Add a configuration option to the handler plugins that allows for the manual addition of labels to metrics passing through that handler. This would allow us to specify a label containing the address in the config file. The drawback here is that we (or the SGO) must manually specify this new label

3. Have the socket plugin append bytes indicating the path of the socket to the message. In this scenario, the handlers must understand the protocol and extract the bytes before parsing the json. The handlers can add a label with the bytes. The drawback again is that a new protocol must exist between the transport and handlers, however unlike #1 this protocol can be made the standard internally and all transports/handlers must abide by it.

4. Create a standard for transport plugins to pass metadata information to handlers on startup. For example, the socket plugin could pass it's path configuration to the handlers so that the handlers can add it as a label to all incoming metrics and events. 

#4 is probably my favorite option.

Issue #57, opened 2021-06-24
sg-core crashing during high volume tests


csibbitt:
While running a test where I drive 100M logs into the system as fast as possible (via the amqp sg-bridge), I was able to reliably crash the sg-core with this stack

```
2021-06-24 16:53:00 [INFO] storing events and(or) logs to Elasticsearch. [plugin: elasticsearch, url: https://elasticsearch-es-http.service-telemetry.svc.cluster.local:9200]
2021-06-24 16:53:00 [INFO] socket listening on /tmp/smartgateway [plugin: socket]
panic: interface conversion: interface {} is nil, not []string

goroutine 175109 [running]:
github.com/infrawatch/sg-core/plugins/application/elasticsearch.(*Elasticsearch).ReceiveEvent(0xc0002d1110, 0xc00138cab0, 0x21, 0x41d8352e1bc00000, 0x2, 0xc0013f6280, 0xf, 0x2, 0xc0013b3260, 0x0, ...)
        /go/src/github.com/infrawatch/sg-core/plugins/application/elasticsearch/main.go:96 +0x87f
github.com/infrawatch/sg-core/pkg/bus.(*EventBus).Publish.func1(0xc00138cab0, 0x21, 0x41d8352e1bc00000, 0x2, 0xc0013f6280, 0xf, 0x2, 0xc0013b3260, 0x0, 0xc001390ca0, ...)
        /go/src/github.com/infrawatch/sg-core/pkg/bus/bus.go:35 +0x62
created by github.com/infrawatch/sg-core/pkg/bus.(*EventBus).Publish
        /go/src/github.com/infrawatch/sg-core/pkg/bus/bus.go:34 +0xbd
```

csibbitt:
https://bugzilla.redhat.com/show_bug.cgi?id=1975951

PR #67, opened 2021-07-12
ES memory write race fix and performance improvement

paramite:
This patch is fixing #57 and:
- allows sending multiple bulk writes in parallel
- also avoid sending whole records via channel in elasticsearch app
- adds pure go (no cgo wrapper) amqp1 transport, which is more performant than with using official qpid C wrapper


plugins/application/elasticsearch/main.go
```diff
@@ -126,16 +126,20 @@ func (es *Elasticsearch) Run(ctx context.Context, done chan bool) {
...
-		case dumped := <-es.dump:
-			if err := es.client.Index(dumped.index, dumped.record, es.configuration.BulkIndex); err != nil {
-				es.logger.Metadata(logging.Metadata{"plugin": appname, "event": dumped.record, "error": err})
-				es.logger.Error("failed to index event - disregarding")
-			} else {
-				es.logger.Debug("successfully indexed document(s)")
+	for i := 0; i < es.configuration.IndexWorkers; i++ {
+		es.logger.Metadata(logging.Metadata{"plugin": appname, "worker-id": i})
+		es.logger.Debug("spawning ES API worker")
+		for {
```
csibbitt, line 136: As discussed, probably missing a `go func{ }` around this for loop

.github/workflows/integration.yml
```diff
@@ -68,7 +68,7 @@ jobs:
           done
           docker logs elastic
           echo "========================== loki ========================="
-          while ! curl -sX GET "http://127.0.0.1:3100/loki/api/v1/ready"
+          while ! curl -sX GET "http://127.0.0.1:3100/ready" | grep -q "^ready$"
```
csibbitt, line 71: Did this commit slip in by mistake? I don't object to it, it just seems out of place.

csibbitt:
Hey @paramite thanks for getting to the bottom of this. I see in a commit comment that you did indeed find problems with concurrent.Map. I'm going to re-read it now to see if I can spot it out of interest.

Just a reminder that force-push is generally a bad idea especially for github. It makes the flow of commits vs. comments illogical, some commits disappear which makes things even harder to follow, and my local branch becomes invalid so I need to delete it and re-pull to get back in sync. In short - please try to avoid force-push. It's okay to rewrite history in private, but once it's been published, you should seek to build upon history, not rewrite it.

plugins/application/elasticsearch/main.go
```diff
@@ -52,15 +52,16 @@ type Elasticsearch struct {
 	configuration *lib.AppConfig
 	logger        *logging.Logger
 	client        *lib.Client
-	buffer        *concurrent.Map
+	buffer        map[string][]string
+	bufferMutex   sync.RWMutex
```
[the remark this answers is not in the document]
pleimer: It looks like he is wrapping the whole write process in the lock rather than just the map. There must have been writes from other go routines happening to the buffer between the check if the index already existed (L92) and writing to the buffer later in the ReceiveEvent() function. This wouldn't necessarily be a problem with the concurrent map but rather how it was being used. That's just my thoughts looking at it, maybe Martin can explain further

plugins/application/elasticsearch/main.go
```diff
@@ -126,21 +130,34 @@ func (es *Elasticsearch) Run(ctx context.Context, done chan bool) {
...
-				es.logger.Error("failed to index event - disregarding")
-			} else {
-				es.logger.Debug("successfully indexed document(s)")
+	wg := sync.WaitGroup{}
+	for i := 0; i < es.configuration.IndexWorkers; i++ {
+		es.logger.Metadata(logging.Metadata{"plugin": appname, "worker-id": i})
+		es.logger.Debug("spawning ES API worker")
+		wg.Add(1)
+
+		go func(es *Elasticsearch, ctx context.Context, wg *sync.WaitGroup, i int) {
```
pleimer, line 139: interesting do you see a significant performance improvement using a worker pool?

pleimer (approved):
Looks good to me. Granted, generally worker pools aren't necessary with goroutines since they are so inexpensive to start-up and the runtime manages the amount that are created.

pleimer:
In fact you could just write to ES inside the ReceiveEvent() function since that is already using it's own goroutine from the bus. It's kinda unnecessary to send it to Run() via channel before posting to ES. This may give even better performance. Just my speculation, I'm still fine with this patch

plugins/application/elasticsearch/main.go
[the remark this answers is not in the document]
csibbitt: Hmmm, I'm still not sure I understand how that caused crashes, but upon re-reading with your explanation in mind, I do think I've spotted where logs were being lost! Between L96 and L102 in the old code; if two routines perform this at the same time, one of their results will be overwritten. This probably accounts for the losses I was seeing at high rates.

csibbitt:
> Looks good to me. Granted, generally worker pools aren't necessary with goroutines since they are so inexpensive to start-up and the runtime manages the amount that are created.

If I understand correctly, you may be suggesting that instead of starting 3 goroutines that each wait on channel and call `es.client.Index` in response, it may be just as safe (and likely more performant) to have just the single channel reader and simply call `es.client.Index` itself inside a goroutine. @paramite has mentioned that the bottleneck now seems to be emptying memory into ES itself (bufferSize can help, but only to a point I guess). Do you think we can just let the go runtime decide how many workers is optimal? I'd certainly be interested in trying that to see how it works out.

csibbitt (approved):
I'm okay with merging this code, but if Paul's suggestions can lead to a simpler code, fewer tunable parameters, and more performance, then I think they are worth trying before we consider this work complete.

pleimer:
> > Looks good to me. Granted, generally worker pools aren't necessary with goroutines since they are so inexpensive to start-up and the runtime manages the amount that are created.
> 
> If I understand correctly, you may be suggesting that instead of starting 3 goroutines that each wait on channel and call `es.client.Index` in response, it may be just as safe (and likely more performant) to have just the single channel reader and simply call `es.client.Index` itself inside a goroutine. @paramite has mentioned that the bottleneck now seems to be emptying memory into ES itself (bufferSize can help, but only to a point I guess). Do you think we can just let the go runtime decide how many workers is optimal? I'd certainly be interested in trying that to see how it works out.

Basically, but I was suggesting re-using the goroutine already spawned by the "bus" when it calls ReceiveEvent() here: https://github.com/infrawatch/sg-core/blob/23ade06b40b9e1e3eb138b9b9b8aa34f9f54cc46/pkg/bus/bus.go#L34.  I was suggesting placing the post to ES in the ReceiveEvent() function, making it so you would never need to create a channel to send events anywhere at all

But yeah that all depends on the configuration of Elasticsearch and how many connections it can handle. It might overwhelm ES if there are too many goroutines posting to it at the same time.

plugins/application/elasticsearch/main.go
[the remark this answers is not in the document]
pleimer: It was crashing here: https://github.com/infrawatch/sg-core/blob/23ade06b40b9e1e3eb138b9b9b8aa34f9f54cc46/plugins/application/elasticsearch/main.go#L96

saying that the type of data at that index in the buffer was not a []string{} value, but a nil. It could be that another goroutine overwrote the value at that index before this type assertion happened. I'm not sure exactly where that would happen, just speculating

paramite:
Hmm, interesting idea Paul. Will play with it in my lab env and in case update this PR. But anyway. This PR at least fixes the panics Chris was experiencing.

paramite:
That was massive performance improvement indeed Paul!

csibbitt:
> That was massive performance improvement indeed Paul!

Sweet. I look forward to re-testing this!  Any sign of problems with too many connections to ES? Are you counting messages into rsyslog == messages into ES ?

paramite:
Hmm, so I'm tailing two log files and adding 2M records to both and after only following gets to the ES:

$ curl -X GET "http://127.0.0.1:9200/sglogs-lenovo_p720_rdo_13.2021.07.15/_count"
{"count":1381952,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0}}

That's not a good number .. sigh

paramite:
Initially I thought that log record loss is happening somewhere in the elasticsearch app and the ES itself, but the print plugin routed to a file reports same number as indexed documents in ES. We are probably loosing data in the sg-core internals somewhere.

csibbitt:
@paramite @pleimer Is there any reason we should not merge this? I see that it doesn't entirely fix the lost logs problem, but AFAICT it's an improvement, right? I'd like to merge this if possible, but since it's a bit stale I don't want to hit the button without a final check with you guys.

vyzigold:
LGTM, the lost logs might have been dropped by the sg-bridge. If @paramite didn't see any lost logs with the quick patch I made to the sg-bridge like two weeks ago, then we should be good with log loss on sg-core side.

leifmadsen:
@paramite if you have no objections to this, I think the consensus is to merge this down.

paramite:
I need to update this a bit and then we can merge, if reviews will be positive.

.github/workflows/tests.yml
```diff
@@ -34,6 +34,8 @@ jobs:
           skip-build-cache: true
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: v1.33
+          # golangci-lint command line arguments.
+          args: --exclude=SA4008
```
leifmadsen, line 38:
```suggestion
          # golangci-lint command line arguments. https://staticcheck.io/docs/checks#SA4008 (The variable in the loop condition never changes, are you incrementing the wrong variable?)
          args: --exclude=SA4008
```

leifmadsen:
Please don't force push over branches. Multiple commits is ok. I often use the commit notes when doing GA releases to make sure I capture changes in the release notes.

paramite:
Leif I'm keeping all relevant changes in relevant commits, but when I need to remove a commit, then it does not make sense to revert it. Instead I remove it and force push. Right now I'm blindly trying to make appstream rsyslog work in the GH action. If I would leave all my trial/error commits, then the PR will be a mess.

paramite:
Yay! CI fixed, cleaned the mess and this is ready for review now.

paramite:
And yes, with Jaromir's sg-bridge patch or without using sg-bridge at all the log loss is 0%

plugins/application/elasticsearch/main.go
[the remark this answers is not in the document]
paramite: Well the problem was that using concurrent.Map's method the data at index are locked only in the method itself, so yes, sometimes it happened that one goroutine fetched the data via Get and before it could work with that data other gouroutine deleted the index in map (because the map was not locked any more). That being said, concurrent.Map was not much ready for concurrency and I did not realize that when I wanted to use it for that purpose.

plugins/application/elasticsearch/main.go
paramite: Fixed.

csibbitt:
> ...force push. Right now I'm blindly trying to make appstream rsyslog work in the GH action. If I would leave all my trial/error commits, then the PR will be a mess.

1) We merge via "Squash and merge" so the main branch will only include a single commit rolling up your PR. I don't think 14 commits + 7 force push messages is any less messy than 21 commits when it comes to the PR screen itself. In fact it would consume less screen space. Heck, others might even learn something by watching you hack against the CI.

2) The solution to hacking against the CI without polluting your PR is not to use force-push, but instead create a second PR branched from the real one, hack on it with impunity, and then bring over the commit that works. Force pushing is an inconvenience to  anyone other than the one doing it (for the reasons I mentioned in July), so it's plainly inconsiderate, especially after multiple people have asked it to stop and provided reasons. I vote we disable on all branches since it promotes dangerous bad habits and provides nothing in return.

plugins/application/elasticsearch/main.go
```diff
@@ -126,21 +132,66 @@ func (es *Elasticsearch) Run(ctx context.Context, done chan bool) {
...
+					}
+					delete(es.buffer, index)
+				}
+				es.bufferMutex.Unlock()
 			}
 		}
+	}(es, ctx, &wg)
+
+	// spawn index workers
+	for i := 0; i < es.configuration.IndexWorkers; i++ {
```
csibbitt, line 167: I'm confused by the history here. On July 13th Paul suggested eliminating the worker pool by re-using the goroutine already spawned by the call ReceiveEvent(). On July 14th you said "That was massive performance improvement indeed Paul!"  but I still see the workers code here. Did this approach need to be adandoned?

atyronesmith (approved):
Reviewed the concurrent map access code.  Looks good.

PR #68, opened 2021-07-29
Fix events arriving out of order to app plugins

vyzigold:
This PR ensures events are processed in the order they arrived on the event bus by the application plugins.

I found out, that because the event bus runs the application plugin receiving functions inside a goroutine, sometimes events can be processed by the application plugins in a different order, than how they arrived on the bus. This is really bad especially for the Loki application plugin, because Loki requires the timestamps of logs to be increasing (last_log_timestamp <= new_log_timestamp). Because of this, sometimes a log sent to Loki would have an older timestamp, than the last log sent to Loki, which would cause "entry out of order" errors.

I wonder, if this behavior is ok for other storages we use (maybe we want to do a similar thing for metrics too). I'm not sure, if and how much, waiting for previous event to finish processing, could affect performance.

### DIFF
--- a/pkg/bus/bus.go
+++ b/pkg/bus/bus.go
@@ -17,6 +17,7 @@ type EventPublishFunc func(data.Event)
 type EventBus struct {
 	subscribers []EventReceiveFunc
 	rw          sync.RWMutex
+	wg          sync.WaitGroup
 }
 
 // Subscribe subscribe to bus
@@ -31,10 +32,13 @@ func (eb *EventBus) Publish(e data.Event) {
 	eb.rw.RLock()
 
 	for _, rf := range eb.subscribers {
-		go func(rf EventReceiveFunc) {
+		eb.wg.Add(1)
+		go func(rf EventReceiveFunc, wg *sync.WaitGroup) {
+			defer wg.Done()
 			rf(e)
-		}(rf)
+		}(rf, &eb.wg)
 	}
+	eb.wg.Wait()
 	eb.rw.RUnlock()
 }
 


vyzigold:
I was looking into where are we loosing logs yesterday. I tried sg-core with and without this patch. I got a lot more amqp_overruns in sg-bridge with this patch (i run sg-bridge with the --block flag), which looks to me like sg-core got noticably slower with this. I'd like to take some measurements today.

pleimer:
Very interesting find.

For metrics, the timestamp created by collectd is used to store the metric into Prometheus, so whether or not a metric arrives slightly out of order does not affect the order it is stored in Prometheus. Side note, this could hinder Prometheus performance, but right now Prometheus is nowhere near the bottleneck in our system. In Elasticsearch, order of timestamps arriving doesn't matter either.

By adding a WG to force goroutines to execute in order, you are essentially re-creating the functionality of a channel. You may see slightly better performance if you utilize those instead. You may not, but it's probably worth testing.

I originally did performance testing for metrics with a bus based on buffered channels but reverted to the goroutine strategy because it was much more performant. It may be that we should change the event bus functionality only and leave the metric bus as is, but document for application plugin developers that messages may arrive out of order

vyzigold:
So I tried to test the speed of this, but I don't think the results are really saying anything unless I test with ES directly (for Loki we need this PR and the application plugin feeds the data directly into a channel, so the speed should be similar). I wanted to test just the bus, so I wanted sg-core to do as little additional work as possible, so I chose the dummy-logs transport plugin and print application plugin to print into a file and just counted how many lines are inside the file, when all the lines were there, I stopped sg-core and measured the time. You can see my testing setup [here](https://github.com/vyzigold/sg-core/tree/bus-testing). I run the "stop" script in the background and then run `time ./sg-core`.
I run the tests multiple times, the results were:
without this PR:
./sg-core -config config.yaml  1033.43s user 68.91s system 367% cpu 5:00.32 total
./sg-core -config config.yaml  1034.51s user 66.30s system 383% cpu 4:46.84 total
./sg-core -config config.yaml  1044.94s user 68.87s system 367% cpu 5:03.06 total
./sg-core -config config.yaml  1037.47s user 63.69s system 377% cpu 4:51.92 total

with this PR:
./sg-core -config config.yaml  510.30s user 76.39s system 211% cpu 4:37.77 total
./sg-core -config config.yaml  496.79s user 73.57s system 212% cpu 4:28.09 total
./sg-core -config config.yaml  488.12s user 73.71s system 210% cpu 4:26.55 total

with channel:
./sg-core -config config.yaml  491.14s user 62.33s system 205% cpu 4:29.80 total
./sg-core -config config.yaml  501.29s user 63.78s system 205% cpu 4:35.23 total

With the PR the sg-core was actually faster probably because it took longer for the fs to synchronize the access to the file, when more goroutines tried to write to it. With channel the speed was comparable, to the speed with this PR, but channel doesn't allow reads from multiple application plugins.

pkg/bus/bus.go
```diff
@@ -31,10 +32,13 @@ func (eb *EventBus) Publish(e data.Event) {
 	eb.rw.RLock()
 
 	for _, rf := range eb.subscribers {
-		go func(rf EventReceiveFunc) {
+		eb.wg.Add(1)
+		go func(rf EventReceiveFunc, wg *sync.WaitGroup) {
+			defer wg.Done()
```
csibbitt, line 37: Doesn't this essentially eliminate any efficiency gained by running the EventReceiveFunc() as a goroutine?

vyzigold: Yes and this is the reason I'm concerned about performance.

**Other ways to solve this would be:** 
Count the logs in the log handler plugin (assign an increasing number to each log) and then sort the logs in loki application plugin, which looks to me like a really bad solution.

Just create a third bus for logs / events, that need this kind of synchronization

Make this code conditional and add a configuration option for application plugins (or somewhere else) and if that option is specified for any of the plugins subscribed to the bus, then this code is used, without it we don't wait for anything.


Now that I think about it, I probably like the last option the most (maybe even more than this)

vyzigold: @pleimer @paramite what do you think?

pleimer: That last option might be a good way to at least unblock logging with Loki for now, and so this patch doesn't interfere with https://github.com/infrawatch/sg-core/pull/67.

In the long run I think we will need to revisit this bus architecture and optimize/enforce order a little better. That might take some refactoring of the core.

pleimer: It should only block functions that are writing to the same subscriber, however. Even though we don't run multiple application plugins in practice, if we did, it's important that writes to one not be blocked by writes to another.

csibbitt: I agree with making this conditional upon some flag that we would only supply in the sg pods that send to loki. I don't think enforcing order is important for most systems, and in fact it would be nice to find out if it's actually possible to relax this in loki (I have some guesses as to perhaps why not, but it's worth a look, I guess)

csibbitt: > That last option might be a good way to at least unblock logging with Loki for now, and so this patch doesn't interfere with #67.

Can we just merge #67 ASAP or is it waiting on anything else? (302 Redirect responses to that PR please)

vyzigold: As far as I know, this and #67 are totally independend. So you don't need to take this PR into account when merging #67. Even if there is some conflict, which I don't see, I plan to make this code conditional as we agreed, so I'll solve any issue when working on that.

@csibbitt There currently isn't any way to relax this in Loki. The requirement is there because of the way Loki stores data (I think). This is currently probably the most often discussed issue in the Loki community and there are some plans to remove this requirement for a little decrease in performance, but I don't know when this feature gets implemented (it could easily be a year).

csibbitt: I kinda expected to hear that. Thanks for looking into it!

pleimer: The optimization done in #67 depended on posts to ES happening (basically) asynchronously. Blocking goroutines would reduce the effect of that optimization, that's what I mean by interfering with that PR.

@csibbitt it's all been approved, I think Martin was doing more testing on that, not sure what the results are

vyzigold:
I added a configuration option "blockEventBus" and when it's set to true, the blocking code is used. I thought about Paul's suggestion to block each application plugin separately, but I'm not sure what would the best way to do this be. I might take a look at this again when we revisit the bus architecture. And as I wrote yesterday we might not even need a blocking bus then. Example configuration to use this would be:
```
pluginDir: /tmp/plugins
logLevel: debug
blockEventBus: true
transports:
    - name: dummy-logs
      handlers:
      - name: logs
        config:
            timestampField: "@timestamp"
            messageField: "message"
            severityField: "severity"
            hostnameField: "host"
applications: 
  - name: print
```

pleimer (changes requested):
It's important that writes to separate applications do not block each other, otherwise this patch is going to cause some pain if we decide to consolidate smart gateways.

You could do something like the following (this is entirely untested code):
```go
// make a new type for each publisher added that we can use to synchronize publisher functions
type blockingPublisher struct {
    rf EventREceiveFunc
    sync.RWMutext
}

func (bp *blockingPublisher) publish(e data.Event) {
    bp.RLock()
    bp.rf(e)
    bp.RUnlock() // defer is actually actually pretty slow, better to free up goroutines sooner than later
}

type EventPublishFunc func(data.Event)

// BlockingEventBus blocking bus for data.Event type
// sg-core would load this object rather than the regular bus when blocking is configured
type BlockingEventBus struct {
    sub_index   int // sub_index keeps track of subscriber groups in proceeding map
    subscribers map[int]*lockablePublisher
    rw          sync.RWMutex
    wg          sync.WaitGroup
}

// Subscribe subscribe to bus
func (eb *BlockingEventBus) Subscribe(rf EventReceiveFunc) {
    eb.rw.Lock()
    defer eb.rw.Unlock()

    eb.subscribers[eb.sub_index] = &lockablePublisher{
        rf: rf,
    }
    eb.sub_index++
}

// Publish publish to bus
func (eb *SyncEventBus) Publish(e data.Event) {
    eb.rw.RLock() // use a write lock instead of read to garauntee 
    for pIndex := 0; pIndex < eb.sub_index; pIndex++ {
        go func(rf EventREceiveFunc) {  // if there are too many 
            eb.subscribers[pIndex].publish(e)
        }
    }
    eb.rw.RUnlock()
}
```

There's actually a bigger problem with the work groups and suggestion I posted above: neither actually guarantee FIFO order. If logs are coming in faster than they can be consumed, there is no guarantee that the first goroutine to arrive will actually actually acquire the lock: https://stackoverflow.com/questions/51140111/is-sync-mutex-lock-a-fifo

We can guarantee order with a buffered channel (just a FIFO queue) until the queue fills up, then it's no longer guaranteed anymore. There is still some cost associated with channels, but it may be a good interim solution. Either way, some stress testing needs to be done to figure how hard we can push it until it starts failing

vyzigold:
@pleimer Thank you for the suggestion. I see the FIFO ordering problem you mentioned in your suggestion, but I think my solution should actually guarantee FIFO order, unless I missed something, each write to the event bus (in my solution) blocks the whole sg all the way to reading from socket and that should be FIFO as well (sg-bridge as far as I know is FIFO as well). I know this isn't ideal, but from my testing so far and from all of the issues with log ordering in the Loki community, I think we really should try to guarantee FIFO ordering if we want to write to Loki (If we don't guarantee FIFO ordering, there is probably no point in even trying to use Loki currently). I don't see a way to guarantee FIFO ordering unless we block everything all the way to the socket and at that point, it probably doesn't make sense to try to not block writes to separate applications (the faster applications would just don't have any data to process, because the socket plugin shouldn't give them anything).

I'd suggest to just use the blocking only for Loki and keep the logging sg separate (we need a different sg-bridge configuration for logs anyway infrawatch/smart-gateway-operator#91). The Loki community is currently working on dropping the ordering requirement, so in the future, if we want to consolidate smart gateways and Loki doesn't require the logs strictly ordered, we can revert this PR and just use the current implementation of the event bus

paramite (approved):
Legit, but I did not tested this though. Will do so and get back here.

pkg/bus/bus.go
paramite: Another option here is to re-timestamp logs according to arrival time to loki app. Question is if the differences between rsyslog timestamp and sg-core timestamp would be a problem?

pleimer:
Thanks the explanation @vyzigold . You're right there really isn't any other way to guarantee order and that none are dropped without blocking everything. Some may be dropped in sg-bridge after a while, but at  that point we would need to start discussing routing to multiple sgs.

pleimer (approved):

PR #69, opened 2021-08-17
Update apputils and fix dummy-logs

vyzigold:
This PR makes sg-core use the newest version of apputils.
Summary of changes in apputils:

- Ability to specify Loki tenant
- Enhanced Loki error message
- Ability to get information about processes

Only affected part by these changes is the Loki plugin (the process info isn't used anywhere). I made the required change to make the Loki plugin compatible. The Loki plugin doesn't use tenants right now and "fake" is the default tenant for Loki.

I also fixed the dummy-logs transport plugin, which wasn't working.


paramite:
Unit tests fail on loki readiness check, which will be fixed in another PR. Legit.

PR #70, opened 2021-08-26
Allow app plugin to be data producers

paramite:
Currently application plugins can only listen on internal buses.
This patch enables apps to also produce messages on internal metric
and event bus.

Partially-implements: STF-519


paramite:
Will reopen with proper branch name.

PR #71, opened 2021-09-01
Minor fixes

paramite:
- do not start application when error happens during initialization
- display also event message in print app


paramite:
Merging since there is no way this can break something

PR #72, opened 2021-09-20
sg-agent necessary changes

paramite:
This PR adds necessary changes required by sg-aget to work correctly.

 - possibility for application plugins to emit events in internal event bus.
 - possibility to register custom configuration validators


paramite:
This is rotting here for a long time already.dfdfdf

PR #73, opened 2021-10-07
Add logs to generator

vyzigold:
This adds a possibility to generate rsyslog log messages. I left metrics generation as a default behavior, logs are generated only when using the -l option.

### DIFF
--- a/generator/amqp_snd_th.c
+++ b/generator/amqp_snd_th.c
@@ -66,7 +66,37 @@ char *CD_MSG3 = ", \"interval\": 1,\"host\": \"";
 char *CD_MSG4 = "\", \"plugin\": \"";
 char *CD_MSG5 = "\", \"plugin_instance\": \"pluginInst0\",\"type\": \"type0\",\"type_instance\": \"typInst0\"}";
 
-static char *build_mesg(app_data_t *app, char *time_buf) {
+char *RSYSLOG_MSG1 = "{\"@timestamp\":\"";
+char *RSYSLOG_MSG2 = "\", \"host\":\"";
+char *RSYSLOG_MSG3 = "\", \"severity\":\"5\", \"facility\":\"user\", \"tag\":\"tag1\", \"source\":\"some-source\", \"message\":\"a log message from generator'\", \"file\":\"\", \"cloud\": \"cloud1\", \"region\": \"some-region\"}";
+
+static char *build_log_mesg(app_data_t *app, char *time_buf) {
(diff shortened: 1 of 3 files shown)


generator/amqp_snd_th.c
```diff
@@ -66,7 +66,37 @@ char *CD_MSG3 = ", \"interval\": 1,\"host\": \"";
...
+char *RSYSLOG_MSG3 = "\", \"severity\":\"5\", \"facility\":\"user\", \"tag\":\"tag1\", \"source\":\"some-source\", \"message\":\"a log message from generator'\", \"file\":\"\", \"cloud\": \"cloud1\", \"region\": \"some-region\"}";
+
+static char *build_log_mesg(app_data_t *app, char *time_buf) {
+    int msg_buf_size = sizeof(app->MSG_BUFFER);
+    char *p = app->MSG_BUFFER;
+
+    for (int i = 0; i < app->num_cd_per_mesg;) {
+        p = memccpy(p, RSYSLOG_MSG1, '\0', msg_buf_size);
+        p--;
+        p = memccpy(p, time_buf, '\0', msg_buf_size);
```
csibbitt, line 80: This is likely not a problem given the relatively static sizes involved here, but I think using `msg_buf_size` repeatedly as the count parameter to memccpy() can lead to overflows. Untested, but I believe you need something more like `msg_buf_size - strlen(p) - 1` on the memccpy()s after L78

csibbitt (changes requested):
This is probably okay as is, but I think the memccpy()s should be adjusted for correctness in case someone copies this code elsewhere or expands it.

generator/amqp_snd_th.c
csibbitt: Ha! So the bad example is already being copied around. :) 

Yes, if you wouldn't mind fixing it in both places and retesting to make sure we're not both wrong, that would be excellent!

csibbitt (approved):

generator/amqp_snd_th.c
atyronesmith: Good catch.  Yes, the comparison should be just > not >= for (p+msg_len) > end

atyronesmith: After looking at this again...  It's possible to a partial log message into the buffer.  It's almost that we need to render the total string and then see if that fits in the remaining buffer space.

atyronesmith (approved):
lgtm

Need to watch for partial messages.

PR #74, opened 2021-10-26
Increase reading buffer size on large messages 

paramite:
When larger than MaxBufferSize messages are being sent, the whole message processing
is failing badly. This can easily happen when socket plugin is used in Ceilometer metrics
or events processing. This patch makes it to increase buffer size, when large message is detected.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2016460

### DIFF
--- a/plugins/transport/socket/main.go
+++ b/plugins/transport/socket/main.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"net"
 	"os"
 	"time"
@@ -14,6 +13,11 @@ import (
 	"github.com/infrawatch/sg-core/pkg/config"
 	"github.com/infrawatch/sg-core/pkg/data"
 	"github.com/infrawatch/sg-core/pkg/transport"
+)
+
+const (
+	initBufferSize = 65535
+	maxBufferSize  = 16384
 )
 
 var (
@@ -87,17 +91,32 @@ func (s *Socket) Run(ctx context.Context, w transport.WriteFn, done chan bool) {
 	}
 	skt.Close()
 
-	s.logger.Infof("socket %s bind successful", laddr.Name)
-	go func() {
+	s.logger.Infof("socket listening on %s", laddr.Name)
+	go func(initBuffSize, maxBuffSize int64) {
+		msgBuffer := make([]byte, initBuffSize)
 		for {
-			var buff bytes.Buffer
-			n, err := io.Copy(&buff, pc)
-			if err != nil {
-				s.logger.Errorf(err, "failed reading message")
-			} else {
-				s.logger.Debugf("read message of size %d bytes", n)
+			n, err := pc.Read(msgBuffer)
+			if err != nil || n < 1 {
+				if err != nil {
+					s.logger.Errorf(err, "reading from socket failed")
+				}
+				done <- true
+				return
 			}
 			msg := buff.Bytes()
+
+			// whole buffer was used, so we are potentially handling larger message
+			// and have to increase buffer size
+			if n == len(msgBuffer) {
+				s.logger.Warnf("insufficient buffer size: %dB", len(msgBuffer))
+				newSize := int64(len(msgBuffer)) + initBuffSize
+				if newSize <= maxBuffSize {
+					msgBuffer = make([]byte, newSize)
+					s.logger.Infof("increased buffer size: %dB", newSize)
+				} else {
+					s.logger.Warnf("reading buffer insufficient, but cannot increase size anymore")
+				}
+			}
 
 			if s.conf.DumpMessages.Enabled {
 				_, err := s.dumpBuf.Write(msg)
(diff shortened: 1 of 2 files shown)


plugins/transport/socket/main.go
```diff
@@ -87,18 +91,33 @@ func (s *Socket) Run(ctx context.Context, w transport.WriteFn, done chan bool) {
...
+					s.logger.Errorf(err, "reading from socket failed")
+				}
+				done <- true
+				return
 			}
 			msg := buff.Bytes()
 
+			// whole buffer was used, so we are potentially handling larger message
+			// and have to increase buffer size
+			if n == len(msgBuffer) {
```
pleimer, line 109: Will this drop one message before it increases the size?

paramite: Yes, unfortunately. The current content is unusable anyway an re-read with larger buffer does read next message instead the whole message.

paramite: That said, I was not able to make things work without using fixed size buffer and to avoid losing the one message when buffer size increases.

plugins/transport/socket/main.go
```diff
@@ -15,7 +15,10 @@ import (
 	"github.com/infrawatch/sg-core/pkg/transport"
 )
 
-const maxBufferSize = 16384
+const (
+	initBufferSize = 65535
+	maxBufferSize  = 16384
```
csibbitt, line 20: `maxBufferSize` is unused now?

plugins/transport/socket/main.go
pleimer: Yeah that's understandable, this buffer increase functionality would have to happen inside the net library for it to be able to consume the initial message. This may be something we need to note somewhere, I'm not really sure of the best place

pleimer (approved):

plugins/transport/socket/main.go
```diff
@@ -96,6 +104,20 @@ func (s *Socket) Run(ctx context.Context, w transport.WriteFn, done chan bool) {
...
+			// whole buffer was used, so we are potentially handling larger message
+			// and have to increase buffer size
+			if n == len(msgBuffer) {
+				s.logger.Warnf("insufficient buffer size: %dB", len(msgBuffer))
+				newSize := int64(len(msgBuffer)) + initBuffSize
+				if newSize <= maxBuffSize {
+					msgBuffer = make([]byte, newSize)
+					s.logger.Infof("increased buffer size: %dB", newSize)
+				} else {
+					s.logger.Warnf("reading buffer insufficient, but cannot increase size anymore")
```
pleimer, line 116: what issues arise if you try to increase the size more than once?

paramite: More than three times. None besides reading large messages which means performance impact.

plugins/transport/socket/main.go
paramite: line 136

paramite: line 136

csibbitt: It's actually L112 that I was looking for - I deleted my comment after I noticed the subtle name change that was happening. Poured another coffee, too :)

plugins/transport/socket/main.go
atyronesmith: The buffer is initialized to the maximum size for a datagram socket (64k) right?  Does the other size send a single message in multiple datagrams?

atyronesmith: Also, since I believe this is the only conduit / pipe ... make the buffer as large as necessary from the very start should not be an issue.  i.e. buffer size = 64k

paramite:
Ok then, modified the patch to just increase the initial buffer size.

plugins/transport/socket/main.go
csibbitt: @paramite - "Does the other si[d]e send a single message in multiple datagrams?"  Have you observed messages >64k?

plugins/transport/socket/main_test.go
```diff
@@ -0,0 +1,83 @@
...
+		conf: configT{
+			Path: sktpath,
+		},
+		logger: &logWrapper{
+			l: logger,
+		},
+	}
+
+	t.Run("test large message transport", func(t *testing.T) {
+		msg := make([]byte, regularBuffSize)
```
csibbitt, line 44: Not sure this test makes sense at this size anymore

csibbitt (approved):
This change looks okay to me. The test could probably use an additional update, and I'm curious if this 64k message limit is enough for our needs, but it's an improvement either way.

paramite:
It is a valid point to ask if 64k is enough. Sadly I don't know. On my environment I could not make ceilometer send bigger messages than original 16k. The unit test is just testing if messages bigger than 16k are parsed correctly. It's not that much useful as it could be, but I guess having not much useful test is better than no test.

paramite:
I guess I will merge this then unless @atyronesmith is against it.

PR #76, opened 2021-11-01
Use latest tag on Dockerfile image references (#75)

leifmadsen:
Cherry picked from commit 8cfdb44f0295274f2d29d1edb3865414def1be90


leifmadsen:
Not really sure why this is failing the logging test. Before this change I believe the image should have defaulted to `latest` anyways. I believe I remember someone telling me that this test isn't super stable and tends to fail, but that could be another test, or might not be true any more.

@vyzigold @paramite any ideas?

csibbitt (approved):

PR #79, opened 2021-11-11
Pin Golang version to 1.15

paramite:


pleimer (approved):

build.sh
```diff
@@ -11,6 +11,7 @@
 
 base=$(pwd)
 
+GOCMD=${GOCMD:-"go"}
```
pleimer, line 14: Maybe we should make this default to go1.15 since that's what our go.sum is based on?

pleimer: that's fair

csibbitt (approved):
I was experiencing "inconsistent vendoring" problems doing containerized builds on the stable-1.3 branch and found this PR which I figured might help. I checked that my master branch suffers similar problems, I merged this PR (to master), and the problems went away. So I'm merging this patch and will attempt to backport it to stable-1.3 now.

I suspect that the problems on the 1.3 branch are slightly different, but this PR helped me better understand what's going on. Thanks!

PR #81, opened 2022-01-25
Cover metrics part with integration tests

paramite:
This patch tests following data paths:

 - Ceilometer > QDR > sg-core > Prometheus
 - collectd > QDR > sg-bridge > sg-core > Prometheus

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,7 +12,7 @@ env:
   TEST_IMAGE: registry.access.redhat.com/ubi8
   TEST_PORT: "-p 3000:3000"
 
-on: push
+on: [push, pull_request]
 
 jobs:
   metrics:
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -2,23 +2,12 @@ name: Integration testing
 env:
   PROJECT_ROOT: /root/go/src/github.com/infrawatch/sg-core
 
-  QDR_IMAGE: quay.io/interconnectedcloud/qdrouterd:1.15.0
+  QDR_IMAGE: quay.io/interconnectedcloud/qdrouterd:1.17.0
   QDR_VOLUME: "--volume=${{ github.workspace }}/ci/service_configs/qdr:/etc/qpid-dispatch:ro"
   QDR_PORT: "-p 5666:5666"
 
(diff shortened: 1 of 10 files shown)


vyzigold (approved):
Cool, lgtm. I wonder, do you use the amqp plugin for ceilometer data just to test the plugin or does sg run like that in production too?

csibbitt:
Honestly I saw `Ceilometer > QDR > sg-core > Prometheus` this and thought "whhaaaaat?"  I do vaguely remember seeing us add the vcabbage code back in (I see it's a fork that Azure adopted), but I guess since it slipped in over the summer alongside the concurrency fixes that had more of my attention, plus the fact that nothing used it, I forgot all about it! 

Obviously this is great news since using this library was the ultimate plan ever since it was discovered (in an unusable state at the time). Now I want to redo all my performance testing with go-amqp! (but this might not be the right time for that)

I'll make my way through the review, but just wanted to comment on this right away because it closes a long loop for us: https://github.com/infrawatch/telemetry-bench/pull/10#issuecomment-541630105  (I've lost the reference to "...and then I read the trello card "  but it was probably something about A) go-amqp having no receiver support at that time (major problem)   and/or   B) it having been recently abandoned by vcabbage)

Anyways, thanks for this. Review coming soon.

csibbitt (approved):
LGTM, thanks for this!

I went through it pretty closely and didn't find much to comment on. My one concern is that you've duplicated the build process (steps from a Dockerfile are re-implemented in the run_sg.sh and run_bridge.sh scripts), which means we need to maintain them in additional places. Is there a reason you are running UBI8+commands in docker to build the applications, rather than building with the existing Dockerfile and using the resulting images?

paramite:
So If I get you first question correct, then you're aking if the whole integration does not use single build process for used container images. IIUC each job in GH actions runs in separate VM, so if I'm not mistaken I would have to build and save these test images to some image repository. I'm not aware of any other way to share images between the jobs. 

Regarding second question. I'm not using Dockerfile building because that would mean that the integration test will run with master version of sg-core and this is not what we want. We need to run the integration tests with the version of sg-core the tests are running for. Also sg-bridge is being used from source only when there is the branch with same name as currently tested sg-core one.

paramite:
I guess I will merge this now and we can improve in next PR if required. Thanks for reviews!

PR #84, opened 2022-02-01
Increase reading buffer size on large messages  (#74)

paramite:
When larger than MaxBufferSize messages are being sent, the whole message processing
is failing badly. This can easily happen when socket plugin is used in Ceilometer metrics
or events processing.

Resolves: rhbz#2016460

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2016460
(cherry picked from commit f13e24114046d0490f0c5c611a3f83d6cd682aaf)
(cherry picked from commit 4af768f9f708098834ae6de674e64638c7445e0b)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,6 +1,7 @@
 name: Integration testing
 env:
   PROJECT_ROOT: /root/go/src/github.com/infrawatch/sg-core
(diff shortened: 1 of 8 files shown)


leifmadsen (approved):
I'm not worried about coveralls failing here since it doesn't seem to be the most sane one here :)

I wonder if we can make it optional somehow...

PR #91, opened 2022-02-07
ES memory write race fix and performance improvement (#67)

paramite:
* Fix Loki readiness check
* Allow multiple API calls to ES
  - This patch adds possibility to submit multiple requests to ES in paralel
     and hence improving performance.
* Revert "Fix concurrent conflicts in elasticsearch (#53)"
  - Unfortunately concurrent.Map is not ready for concurrency,
     so we need to avoid using it.
  - This reverts commit e61141ab8d2590a7b91d18cd494321af763507bd.
* Fix concurrency in ES client message buffer
  -  "fatal error: concurrent map writes"
      /go/src/github.com/infrawatch/sg-core/plugins/application/elasticsearch/main.go:95
     /go/src/github.com/infrawatch/sg-core/pkg/bus.go:65 calls Application callbacks as a goroutine
     which results in concurrency in ReceiveEvent(). I've added a mutex to protect access.
  - Co-authored-by: Chris Sibbitt <csibbitt@redhat.com>
* Remove unused package
  - concurrent.Map is unused throughout the project and it did not work as expected,
     because values could be overwritten when unlocked after returning from method.
* Add amqp1 transport
* Don't transfer whole message through channels
   - This patch makes channel synchronization use pointers instead of whole messages,
     which improves performance of the elasticsearch app plugin. It also adds mechanism
     to flush index buffer in case of remnants of some batch messages are kept
      for longer time.
* Integration CI fixes
  - simplify rsyslog configuration to work with newer rsyslog
  - stop using ELN repo because of rsyslog-omamqp1
     (plugin is available in appstream)
  - add proper wait time between steps instead of simple sleep
* Update .github/workflows/tests.yml
  - Co-authored-by: Leif Madsen <lmadsen@redhat.com>
* Move golint exclude-rules to proper place

(cherry picked from commit 7d46ba5a90908c30487dc4cea6530afc220ed49a)


leifmadsen (dismissed):

paramite:
Rebased on latest stable-1.3

PR #92, opened 2022-02-08
Stabilize collectd-metrics unit tests

paramite:
This patch avoids testing parsed metrics per metric and instead tests whole slice.
It also joins testing data to the test source code itself.


csibbitt (approved):
LGTM

leifmadsen:
All contained in main_test.go so I'm good with this.

leifmadsen (approved):

PR #95, opened 2022-02-22
Build test plugins for upstream distribution

leifmadsen:
Build the test plugins dummy-* for upstream builds so that when testing
with smart-gateway-operator by default the plugins exist to simulate
workloads without needing a source connected.

Changes will be used as part of the smart-gateway-operator example
CustomResource and allow for additional molecule testing in the future
for that Operator.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -10,7 +10,7 @@ COPY . $D/
 COPY build/repos/opstools.repo /etc/yum.repos.d/opstools.repo
 
 RUN dnf install golang git qpid-proton-c-devel -y --setopt=tsflags=nodocs
-RUN go install golang.org/dl/go1.15@latest && /go/bin/go1.15 download && PRODUCTION_BUILD=true GOCMD=/go/bin/go1.15 ./build.sh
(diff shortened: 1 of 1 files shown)


leifmadsen:
Not sure what is up with the assembly error, but it is unrelated to this change. Note how `dummy-*` plugins are now being built.

```
CONTAINER_BUILD=true ./build.sh
building amqp1.so
building dummy-alertmanager.so
building dummy-events.so
building dummy-logs.so
building dummy-metrics.so
building socket.so
building ceilometer-metrics.so
building collectd-metrics.so
building events.so
building logs.so
building sensubility-metrics.so
building alertmanager.so
building elasticsearch.so
building loki.so
building print.so
building prometheus.so
# github.com/cespare/xxhash/v2
asm: xxhash_amd64.s:120: when dynamic linking, R15 is clobbered by a global variable access and is used here: 00092 (/home/leif/src/github.com/infrawatch/sg-core/vendor/github.com/cespare/xxhash/v2/xxhash_amd64.s:120)	ADDQ	R15, AX
asm: assembly failed
building sg-core for container
```

csibbitt (approved):

PR #98, opened 2022-06-21
fixup/golang linting

leifmadsen:
- go mod tidy
- Test with go mod download for missing deps
- Update golangci-lint action to v3

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,8 +25,8 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.15'
-      #- name: download libraries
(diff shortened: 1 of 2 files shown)


leifmadsen:
I think the issue here was that golangci-lint-action was installing golang 1.18 which was failing on our code base (we seem to target golang 1.15). I noticed in the `v3` version of the action it now longer installs golang automatically and now requires you to use the setup-go action (which we were using, but wasn't resulting in the version of golang that was being targeted).

These changes bump the versions to what was recommended on the README for golangci-lint-action repository, along with action call order and combined cache parameter.

mrunge (approved):
Looks good to me.

PR #101, opened 2022-10-03
Add udp transport to socket plugin

vyzigold:
This adds udp to our current socket transport plugin. The unix socket functionality stays unchanged and is the default behaviour. To use udp socket, add "type: udp" and "url: host:port" to the config file for example like this:

```
pluginDir: /tmp/plugins
logLevel: debug
transports:
    - name: socket
      config:
          url: "127.0.0.1:3060"
          type: udp
      handlers:
      - name: logs
        config:
            timestampField: "@timestamp"
            messageField: "message"
            severityField: "severity"
            hostnameField: "host"
applications: 
  - name: print
    config:
      metricoutput: "/dev/null"
      eventsoutput: "/dev/stdout"
```


plugins/transport/socket/main.go
```diff
@@ -31,7 +32,9 @@ func rate() int64 {
 }
 
 type configT struct {
-	Path         string `validate:"required"`
+	Path         string
+	Type         string
+	Url          string
```
csibbitt, line 37: Thanks to the bot for bringing it up. "127.0.0.1:3060" is not a url since it's missing a scheme and doesn't support a path. "socketaddr" may be more appropriate

plugins/transport/socket/main.go
```diff
@@ -80,17 +81,48 @@ func (s *Socket) Run(ctx context.Context, w transport.WriteFn, done chan bool) {
...
 	}
 	skt.Close()
 
 	s.logger.Infof("socket listening on %s", laddr.Name)
+
+	return pc
+}
+
+func (s *Socket) initUdpSocket() *net.UDPConn {
+	addr, err := net.ResolveUDPAddr("udp", s.conf.Url)
```
csibbitt, line 100: If you don't want to explicitly test `err` (since I assume errors in resolution would be caught on the next line in the ListenUDP call) then you should do `addr, _ := ...`. Handling the error explicitly could result in a better troubleshooting experience, though.

plugins/transport/socket/main.go
```diff
@@ -80,17 +81,48 @@ func (s *Socket) Run(ctx context.Context, w transport.WriteFn, done chan bool) {
...
+
+// Run implements type Transport
+func (s *Socket) Run(ctx context.Context, w transport.WriteFn, done chan bool) {
+	var pc net.Conn
+	if s.conf.Type == "unix" {
+		pc = s.initUnixSocket()
+	} else if s.conf.Type == "udp" {
+		pc = s.initUdpSocket()
+	} else {
+		s.conf.Type = "unix"
```
csibbitt, line 120: What is the purpose of this assignment?

csibbitt (commented):
LGTM, just a couple small suggestions and a minor question

leifmadsen:
Just want to ack that I did look through this, and I don't have any further comments beyond what the bot and Chris has mentioned.

plugins/transport/socket/main.go
```diff
@@ -31,7 +32,9 @@ func rate() int64 {
 }
 
 type configT struct {
-	Path         string `validate:"required"`
+	Path         string
```
paramite, line 35: I believe that one of path or url would be always required. You can achieve that with validate:"required_without=Url" and vice versa

plugins/transport/socket/main.go
paramite: Agree. Then you don't need two config parameters.

plugins/transport/socket/main.go
paramite: Yup having information in log (at least) about issue with socket address would improve debugging.

plugins/transport/socket/main.go
vyzigold: Yes, the socketaddr sounds a lot better. I'm not sure about not needing 2 config parameters. I think it would be a good idea for backwards compatibility of the config file to leave the path as a possible parameter for unix sockets.

plugins/transport/socket/main.go
vyzigold: Good catch, I overlooked that one, I'll add a test for the error there.

plugins/transport/socket/main.go
vyzigold: This was a "last minute" attempt to fix the unit tests, which don't set the s.conf.Type to anything before our meeting yesterday. In order for this to work, that assignment would need to be before the first "if". Sorry for that

plugins/transport/socket/main.go
vyzigold: Thanks, I didn't know about this.

plugins/transport/socket/main.go
leifmadsen: Yes please, otherwise we need to adjust SGO to account for any configuration file changes.

leifmadsen (approved):

leifmadsen:
I wonder if we should even bother with code coverage checks. I wonder if there is a way we can make that more informational rather than "voting" as it makes the testing look odd.

Obviously out of scope for this PR, I just wanted to comment on it while it was being shown.

csibbitt (approved):

leifmadsen:
I'm merging this down so it can be included with STF 1.5 as we'll likely be using this release as part of RHOSP 17.1.

PR #102, opened 2022-10-12
[DNM] Test PR for finding why integration testing fails

leifmadsen:


leifmadsen:
Creating this PR to try and test the integration CI testing which seems to be failing after the most recent merge. Everything was working fine in the #101 PR but after merging, a couple of the integration tests are failing. It's not immediately obvious to me why they might be failing.

leifmadsen:
Well then.. no idea why it's failing on master, but it seems to be passing with and without the call to the `type: unix` so that isn't required. I'm feeling better though that the test is likely a false-positive on `master`.

PR #103, opened 2022-11-28
Add TCP to the socket transport plugin

vyzigold:
This PR adds TCP to the socket plugin, which allows us to receive messages with similar speed to UDP, but reliably. A lot more code changes were required in comparison to UDP, because we need to handle each TCP connection separately.

I also included a few changes to the ceilometer handler plugin, which are required to use TCP and UDP.

Sg should still behave the same (use unix socket), unless it's explicitly specified in the configuration to use UDP or TCP.

Example configuration, which uses this functionality:
```
pluginDir: /tmp/plugins
logLevel: debug

transports:
  - name: socket
    config:
      socketaddr: 192.168.122.13:6420
      type: tcp
    handlers:
      - name: ceilometer-metrics
        config:
          source: tcp

applications:
  - name: prometheus
    config:
        host: 0.0.0.0
        port: 3000
```

The UDP and TCP ceilometer publishers use msgpack format (which is basically a binary JSON) instead of JSON. The ceilometer handler plugin needs to know which format to expect and I don't think there is a way for the transport plugin to tell it which one to expect. I solved this by adding the "source" configuration option to the ceilometer handler plugin. I'm not sure, if "source" with possible values: "tcp", "udp" or "unix" is the best solution. Another idea I had, which makes a bit more sense for a handler plugin configuration, was to call the option "format" with possible values: "json" and "msgpack", but this might be confusing to potential users (they would need to match the "type" and "format" options correctly).


vyzigold:
I discovered, that this code currently has an issue, it crashes after a while when trying to receive from multiple publishers at once. If it doesn't crash, it might lose some metrics.

I double checked my original POC. The POC works. I sent 2.4 million of metrics from 4 publishers running at the same time with 100% reliability. I'll check the code in this PR tomorrow and compare it with the POC. My guess is a race condition somewhere.

plugins/handler/ceilometer-metrics/pkg/ceilometer/ceilometer.go
```diff
@@ -73,6 +74,22 @@ func (c *Ceilometer) ParseInputJSON(blob []byte) (*Message, error) {
...
 
+// ParseInputMsgPack parse blob into list of metrics
+func (c *Ceilometer) ParseInputMsgPack(blob []byte) (*Message, error) {
+	msg := &Message{}
+	metric := Metric{}
+	err := msgpack.Unmarshal(blob, &metric)
+	if err != nil {
+		return nil, err
+	}
+	err = msgpack.Unmarshal(blob, &msg)
```
csibbitt, line 85: This looks suspicious to me. You've already applied the `&` operator above when declaring msg. I've looked at the Unmarshal code and AFAICT it will attempt to decode this as an int. Could this be the source of your missing metrics? I haven't read closely enough to know if this explain the crash, but maybe?

vyzigold: You are right, that it looks a bit suspissious and this line is something, which isn't in my original POC. This'll be the first thing I look at tomorrow. But one thing, this doesn't explain is, that this code works perfectly, when I send metrics to it from only one publisher. The issues start, when I use more publishers at once.

plugins/transport/socket/main.go
```diff
@@ -201,10 +199,13 @@ func (s *Socket) ReceiveData(maxBuffSize int64, done chan bool, pc net.Conn, w t
 		}
 
 		if s.conf.Type == tcp {
-			remainingMsg, err = s.WriteTCPMsg(w, msgBuffer, n)
+			parsed, err := s.WriteTCPMsg(w, msgBuffer, n)
 			if err != nil {
+				s.logger.Errorf(err, "error, while parsing messages")
 				return
 			}
+			remainingMsg = make([]byte, int64(n) - parsed)
```
csibbitt: Okay, this is making more sense. I was reading over this code just a few minutes ago trying to formulate a comment to explain exactly what was bothering me about it. Seems you figured it out by running it more; good job! :)

csibbitt (approved):
This LGTM. I haven't run it, but I examined closely enough to spot a missing `&` and get lost in the (broken at the time) large message processing, which now makes clear sense to me. This is pretty key code, so I'd appreciate more eyes if we can get them.

CC: @paramite @yadneshk @atyronesmith

plugins/transport/socket/main.go
```diff
@@ -115,50 +121,144 @@ func (s *Socket) initUDPSocket() *net.UDPConn {
...
-	}(maxBufferSize)
+	}
+}
+
+// Run implements type Transport
+func (s *Socket) Run(ctx context.Context, w transport.WriteFn, done chan bool) {
+	var pc net.Conn
+	switch s.conf.Type {
+	case udp:
+		pc = s.initUDPSocket()
```
leifmadsen, line 224: Is this named `pc` for historical reasons? Just seems odd with TCPSocket being used for `s.initTCPSocket` and not using `UDPSocket` for `s.initUDPSocket`.

plugins/transport/socket/main.go
```diff
@@ -115,50 +121,144 @@ func (s *Socket) initUDPSocket() *net.UDPConn {
...
+		if pc == nil {
+			s.logger.Errorf(nil, "Failed to initialize socket transport plugin")
```
leifmadsen, line 226:
```suggestion
			s.logger.Errorf(nil, "Failed to initialize UDP socket transport plugin")
```

plugins/transport/socket/main.go
```diff
@@ -115,50 +121,144 @@ func (s *Socket) initUDPSocket() *net.UDPConn {
...
+			return
+		}
+		go s.ReceiveData(maxBufferSize, done, pc, w)
+
+	case tcp:
+		TCPSocket := s.initTCPSocket()
+		if TCPSocket == nil {
+			s.logger.Errorf(nil, "Failed to initialize socket transport plugin")
```
leifmadsen, line 234:
```suggestion
			s.logger.Errorf(nil, "Failed to initialize TCP socket transport plugin")
```

plugins/transport/socket/main.go
```diff
@@ -115,50 +121,144 @@ func (s *Socket) initUDPSocket() *net.UDPConn {
...
+			return
+		}
+		go func() {
+			for {
+				pc, err := TCPSocket.AcceptTCP()
```
leifmadsen, line 239: Still a little confused here, hoping my previous question basically helps answer this bit here because the variable reuse isn't making sense to me currently. I guess my only question here is helping me understand :)

plugins/transport/socket/main.go
```diff
@@ -115,50 +121,144 @@ func (s *Socket) initUDPSocket() *net.UDPConn {
...
+				}
+				go s.ReceiveData(maxBufferSize, done, pc, w)
+			}
+		}()
+	case unix:
+		fallthrough
+	default:
+		pc = s.initUnixSocket()
+		if pc == nil {
+			s.logger.Errorf(nil, "Failed to initialize socket transport plugin")
```
leifmadsen, line 257:
```suggestion
			s.logger.Errorf(nil, "Failed to initialize unix socket transport plugin")
```

plugins/transport/socket/main.go
leifmadsen: Perhaps this isn't even correct because this is the socket transport plugin, regardless of the connection method (`source`?). I might not be making a valid suggestion here, or maybe it should be phrased differently, but it feels like we should differentiate the connection type in the log message for future debugging.

plugins/transport/socket/main.go
```diff
@@ -210,17 +309,17 @@ func (s *Socket) Config(c []byte) error {
 	}
 
 	s.conf.Type = strings.ToLower(s.conf.Type)
-	if s.conf.Type != unix && s.conf.Type != udp {
-		return fmt.Errorf("unable to determine socket type from configuration file. Should be either \"unix\" or \"udp\", received: %s",
+	if s.conf.Type != unix && s.conf.Type != udp && s.conf.Type != tcp {
+		return fmt.Errorf("unable to determine socket type from configuration file. Should be either \"unix\", \"udp\" or \"tcp\", received: %s",
```
leifmadsen, line 313:
```suggestion
		return fmt.Errorf("unable to determine socket type from configuration file. Should be one of \"unix\", \"udp\" or \"tcp\", received: %s",
```

plugins/handler/ceilometer-metrics/main.go
```diff
@@ -71,7 +82,13 @@ func (c *ceilometerMetricHandler) Run(ctx context.Context, mpf bus.MetricPublish
...
 func (c *ceilometerMetricHandler) Handle(blob []byte, reportErrs bool, mpf bus.MetricPublishFunc, epf bus.EventPublishFunc) error {
 	c.totalMessagesReceived++
-	msg, err := c.ceilo.ParseInputJSON(blob)
+	var msg *ceilometer.Message
+	var err error
+	if c.config.Source == "unix" {
+		msg, err = c.ceilo.ParseInputJSON(blob)
+	} else {
+		msg, err = c.ceilo.ParseInputMsgPack(blob)
+	}
```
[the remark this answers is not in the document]
vyzigold: True

plugins/transport/socket/main.go
vyzigold: So what about something like: "Failed to initialize socket transport plugin with type: " and after that would be either TCP, UDP or unix?

plugins/transport/socket/main.go
vyzigold: As I wrote in the previous question. This line returns basically the same thing as the `initUDPSocket` and `initUnixSocket` functions.

plugins/transport/socket/main.go
leifmadsen: Ah gotcha. Thanks for the overview!

plugins/transport/socket/main.go
leifmadsen: That works for me!

leifmadsen (approved):
Just the two minor things I called out, otherwise I'm happy with this (through my limited golang abilities; those better able to review the code should work through this as well).

PR #112, opened 2023-08-23
Added a capability for generating ceilometer messages

vyzigold:
Recently I needed a way to generate custom ceilometer messages. So I added a possibility to generate them with the generator.

While doing that I also removed some tab characters and replaced them with spaces. I'm sorry for the confusion


vyzigold:
Hmm, I see both of the integration tests failed. Does that happen often?

csibbitt (approved):
LGTM. One optional improvement that didn't originate with this PR.

vyzigold:
@csibbitt I liked your suggestion, so I implemented it. The reason why I originally programmed it like that was, that when I added rsyslog messages to the generator, I tried to touch the collectd generation as little as possible. So when running the generator like normal, you would generate collectd messages and if you wanted rsyslog messages, you had to specify a special parameter (-l). But I guess now, with all the other changes, it makes sense and it looks a lot better to add and use the app->collectd variable.

vyzigold:
I redased on top of master, which has fixed CI now. If nobody has any concerns, I'll merge tomorrow.

leifmadsen:
Rebased on latest merged changes to the CI refactoring in order to get this cleared. Thanks!

PR #113, opened 2023-08-25
Fix integration CI

vyzigold:
This should fix the integration CI.

It looks like the sg-core's plugins aren't inside the plugin folder. I see the following errors in the CI output when it's trying to build the plugins:
```
error obtaining VCS status: exit status 128
Use -buildvcs=false to disable VCS stamping.
```

From what I found, go >= 1.18 is adding some version control information to the built binaries and this seems to be failing right now. This PR disables that. I don't see the feature being much useful. If we decide it is useful, we can find another way.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -14,7 +14,7 @@ base=$(pwd)
 GOCMD=${GOCMD:-"go"}
 PLUGIN_DIR=${PLUGIN_DIR:-"/tmp/plugins/"}
 CONTAINER_BUILD=${CONTAINER_BUILD:-false}
-BUILD_ARGS=${BUILD_ARGS:-'-buildvcs=false'}
+BUILD_ARGS=${BUILD_ARGS:-''}
 
 PRODUCTION_BUILD=${PRODUCTION_BUILD:-false}
 if $PRODUCTION_BUILD; then
--- a/build.sh
+++ b/build.sh
@@ -14,6 +14,7 @@ base=$(pwd)
 GOCMD=${GOCMD:-"go"}
 PLUGIN_DIR=${PLUGIN_DIR:-"/tmp/plugins/"}
 CONTAINER_BUILD=${CONTAINER_BUILD:-false}
+BUILD_ARGS=${BUILD_ARGS:-'-buildvcs=false'}
 
 PRODUCTION_BUILD=${PRODUCTION_BUILD:-false}
 if $PRODUCTION_BUILD; then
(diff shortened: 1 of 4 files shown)


vyzigold:
Lets see if this works

vyzigold:
The fix worked, but one of the 4 actions failed because of a timeout on running rsyslog. The timeout was 180s and on the successful run it took a little 150s for rsyslog to start. I increased the timeout to 240s hoping, that it'll improve the CI stability.

csibbitt:
I have a hunch this VCS info could be required for downstream builds (which depend on this build.sh), but  I haven't been able to prove it, yet.

build.sh
```diff
@@ -14,6 +14,7 @@ base=$(pwd)
 GOCMD=${GOCMD:-"go"}
 PLUGIN_DIR=${PLUGIN_DIR:-"/tmp/plugins/"}
 CONTAINER_BUILD=${CONTAINER_BUILD:-false}
+BUILD_ARGS=${BUILD_ARGS:-'-buildvcs=false'}
```
csibbitt, line 17: I think it might be better to default this to '' and set `-buildvcs=false` only during CI for now

csibbitt:
This LGTM but it's worth noting that we build with golang 1.18.10 downstream and have not needed this flag. All else being equal, I'd rather the vcs info is in the binary if possible. So my vote is to disable the feature only where it's failing in CI.

CC: @leifmadsen @compi-migui Heads up on this in case we notice vcs related problems downstream (either because it's still on and it's broken, or we turned it off and it's required)

csibbitt (commented):
This LGTM but it's worth noting that we have been building with golang 1.18.10 downstream and have not needed this flag. All else being equal, I'd rather the VCS info is in the binary if possible, though I see that perhaps the layout is causing problems? Unclear

build.sh
vyzigold: That'll be easy, I'll do that first thing on Monday. Another thing that I thought about was to try `rm -rf .git` before building, which might work based on what I read. But that way I think you also lose the VCS info in the binaries and I think your suggestion will be the most elegant anyway.

csibbitt (approved):

build.sh
leifmadsen: When I first read this I was concerned this approach could break the CI system via GHA, but I was thinking of this bit from sg-bridge: https://github.com/infrawatch/sg-bridge/blob/master/.github/workflows/linter.yml#L10-L29

If a `rm -rf .git/` is required, then it'd be good to make sure that job setup is done in a way where it would wouldn't make sense to have to interact with git after it was done (like in the above linting auto-fix scenario with sg-bridge).

I lack concern now :)

leifmadsen (approved):

PR #117, opened 2023-09-06
Rewrite integration CI

vyzigold:
As we discussed with Leif previously, I rewrote the CI to be closer to how we actually use sg-core.

- The original metrics job is now split into 2. Ceilometer and Collectd have their own job, which makes it easier to see which one of them is failing.
- I switched the transport for ceilometer from the amqp transport plugin to use sg-bridge and the socket transport plugin, just like we do in STF
- I added a new job, which tests sg-core with ceilometer by using the TCP as transport


vyzigold:
I think it might be a good idea to somehow make it, that the jobs don't get executed twice on PR. Because it's getting a little cluttered now with 16 jobs running instead of just 8. I thought it might be solved by changing the "on" parameter in the workflow file, but for example here: https://github.com/openstack-k8s-operators/telemetry-operator/pull/172 The "on" parameter is also `on: [push, pull_request]`, but the CI job runs only once.

vyzigold:
I tried to remove the mysql workaround from the workflow to see, if it's still required. I'll delete the commit if this fails.

vyzigold:
I deleted the commit. Both of the workarounds (mysql, postgres) seem to be required to deploy devstack.

vyzigold:
I'm done trying things, this is ready for review

.github/workflows/integration.yml
```diff
@@ -58,17 +64,111 @@ jobs:
...
-      - name: Start sg-bridge for collectd from container image
+      - name: Start sg-bridge from container image
         if: steps.bridge_branch.outcome != 'success'
         run: |
           docker run --name=sgbridge --network host $BRIDGE_VOLUME -d \
-            $BRIDGE_IMAGE --amqp_url amqp://localhost:5666/$QDR_CHANNEL_COLLECTD \
+            $BRIDGE_IMAGE --amqp_url amqp://localhost:5666/$QDR_CHANNEL \
             --gw_unix=$BRIDGE_SOCKET
-      - name: Start sg-bridge for collectd with same branch
+      - name: Start sg-bridge for with same branch
```
leifmadsen, line 73:
```suggestion
      - name: Start sg-bridge from same branch
```

.github/workflows/integration.yml
```diff
@@ -58,17 +64,111 @@ jobs:
...
         if: steps.bridge_branch.outcome == 'success'
         run: |
           docker run --name=sgbridge --network host $BRIDGE_VOLUME -d -uroot \
-            -e GITHUB_REF -e BRIDGE_SOCKET -e QDR_CHANNEL_COLLECTD -e OPSTOOLS_REPO \
+            -e GITHUB_REF -e BRIDGE_SOCKET -e QDR_CHANNEL -e OPSTOOLS_REPO \
+            --workdir=$(dirname $BRIDGE_SOCKET) \
+            $TEST_IMAGE bash $PROJECT_ROOT/ci/integration/metrics/run_bridge.sh
+      - name: Install collectd
+        run: |
+          sudo apt-get install collectd
```
leifmadsen, line 82: What version of collectd does this install?

.github/workflows/integration.yml
```diff
@@ -58,17 +64,111 @@ jobs:
...
+          docker logs sgcore
+          echo "======================== prometheus =====================" && \
+          docker logs prometheus
+      - name: Validate metrics processing
+        run: |
+          docker run --name=validate -uroot --network host \
+            --volume ${{ github.workspace }}:$PROJECT_ROOT:z --workdir $PROJECT_ROOT \
+            $TEST_IMAGE bash $PROJECT_ROOT/ci/integration/metrics/collectd/run_validation.sh
+#-------------------------------------------------------------------------------
+  ceilometer-metrics-bridge:
```
leifmadsen, line 113: Just noting that further refactoring could happen here using composite actions (https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) in order to remove some of this duplication.

I am not advocating doing it here though, just noting in case a future refactoring was warranted.

.github/workflows/integration.yml
leifmadsen: I see now you've just moved what was already happening up here. Still curious, but this seems fine :)

leifmadsen (approved):
Went through the code and some existing scripts etc to see how this looks, and see some of the interactions. Couple minor questions I was able to answer. Looks good. Thanks!

.github/workflows/integration.yml
vyzigold: I don't have ubuntu on hand right now, but I found this: https://www.ubuntuupdates.org/package/core/jammy/universe/updates/collectd So I guess it could be 5.12?

.github/workflows/integration.yml
vyzigold: Yes I agree. For the metric jobs for example, the first ~5 steps are the same. I was looking into the composite actions but I decided to not implement them. As I understand the documentation, we would either need to create a new repository for the actions or we could have them in this repo, but we would probably need to tag the repository. So that we can use the `uses: action@tag` syntax. I haven't tried this, but after reading the docs and understanding it like this, I decided to not use the composite actions to not create some confusion in the repository tags just for CI deduplication.

.github/workflows/integration.yml
```diff
@@ -58,17 +64,111 @@ jobs:
...
+        run: |
+          echo "${GITHUB_REF#refs/heads/}"
+          git ls-remote --exit-code --heads https://github.com/infrawatch/sg-bridge.git "$(echo ${GITHUB_REF#refs/heads/})"
+      - name: Start sg-bridge from container image
+        if: steps.bridge_branch.outcome != 'success'
+        run: |
+          docker run --name=sgbridge --network host $BRIDGE_VOLUME -d \
+            $BRIDGE_IMAGE --amqp_url amqp://localhost:5666/$QDR_CHANNEL \
+            --gw_unix=$BRIDGE_SOCKET
+      - name: Start sg-bridge for with same branch
```
vyzigold, line 167:
```suggestion
      - name: Start sg-bridge from same branch
```

vyzigold:
Just FYI. The logging CI job just failed. I looked at the logs and it failed to install the golang package, so not our fault right now. I'll restart the job.

leifmadsen:
Looks good to me! Thanks for doing this!

csibbitt (approved):
LGTM

PR #120, opened 2023-10-26
Import master to stable-1.5

leifmadsen:
Import master into stable-1.5 in preparation for STF 1.5.3 release.

- Add TCP to the socket transport plugin (#103)
- Move to golang 1.19 (#104)
- Bump github.com/prometheus/client_golang from 1.11.0 to 1.11.1 (#107)
- Specify handler name in handler config error msg (#109)
- Fix integration CI (#113)
- Rewrite integration CI (#117)
- Bump golang.org/x/sys from 0.0.0-20210630005230-0f9fa26af87c to 0.1.0 (#108)
- Bump gopkg.in/yaml.v3 from 3.0.0-20210107192922-496545a6307b to 3.0.0 (#116)
- Added a capability for generating ceilometer messages (#112)
- Add new fields to sg-core metric messages (#111)
- Add devstack plugin (#118)

PR #121, opened 2023-11-01
[zuul] Add zuul jobs to sg-core

elfiesmelfie:
Depends-On: http://github.com/infrawatch/service-telemetry-operator/pull/512

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,0 +1,8 @@
+
+---
+- project:
+    name: infrawatch/sg-core
+    github-check:
+      jobs:
+        - stf-crc-latest-nightly_bundles
+        - stf-crc-latest-local_build


csibbitt:
The ceilometer integration test failures are a known problem[1] not related to this change. The RDO jobs themselves worked just fine.

[1] https://github.com/infrawatch/sg-core/pull/120#issuecomment-1781417153

.zuul.yaml
```diff
@@ -0,0 +1,8 @@
+
+---
+- project:
+    name: infrawatch/sg-core
+    github-check:
+      jobs:
+        - stf-crc-latest-nightly_bundles
```
csibbitt, line 7: @elfiesmelfie Does this nightly bundles test do anything involving the change, or is it just a side by-side to help us quickly rule out any non-change-related problems?

csibbitt (approved):

.zuul.yaml
elfiesmelfie: It is a side-by-side to help us rule out issues. 

Eventually, this will be run periodically, rather than on a change-by-change basis. However, since there's a lot of change for stf-run-ci at the moment, it's not a bad idea to run the job as often as possible to shake out bugs

PR #122, opened 2023-11-06
[zuul] Use the stf-crc-jobs project template

elfiesmelfie:
Instead of updating the .zuul.yaml file everytime
infrawatch/service-telemetry-operator adds a new job, the project-template can be updated in STO, which will propogate the change to the project and avoid leaving gaps in testing.

Depends-On: https://github.com/infrawatch/service-telemetry-operator/pull/514


leifmadsen:
recheck

elfiesmelfie:
> recheck

Looks like the two expected jobs are running!

elfiesmelfie:
recheck

leifmadsen:
> The coveralls job that is failing seems to be a false failure.
> 
> The file that is flagged as having lower coverage is https://github.com/infrawatch/sg-core/blob/master/plugins/handler/events/pkg/lib/source.go, which has not been modified recently. The test_source.go file was also not modified recently. This probably is due to recenet ceilometer enablement, but is unrelated to this patch.

Yea that happens sometimes. I keep looking periodically for a threshold since it's a bit fuzzy, but nothing shows up. You can pretty much just ignore it failing since it's a very small value.

elfiesmelfie:
The failures are not related to this change

PR #124, opened 2023-11-23
Add support for deploying with docker to devstack

vyzigold:
In upstream openstack CI on older stable branches (for example stable/yoga), we use ubuntu based VMs, which don't yet have the podman package it their package manager and the installation fails. For these distros, we need to use docker instead.

I added a code, that would install docker if it fails to install podman. It also configures docker to be usable by the stack user.

I should note, that I wasn't able to create a ubuntu 20.04 or ubuntu 18.04 VM, which should be the affected ubuntus so I tested with debian 10 instead, which also doesn't have podman. I wasn't able to get the whole devstack deployed though, because debian 10 isn't supported and it failed because of unrelated issues, but the devstack got far enough to execute the `install_container_executable` here. It correctly installed docker and I was able to execute the rest of the steps in this plugin manually without issues.


leifmadsen:
We can ignore rdoproject.org/github-check job here since it isn't covered by CI.

paramite (approved):

PR #126, opened 2023-12-13
Upgrade to golang 1.21

csibbitt:
Result of `go mod tidy -go=1.21`


csibbitt:
I built and tested this locally, and it passed everything except the ES and LOKI tests (which I think is expected since I didn't set up any targets). Looks like I have a bunch more to do for the CI tests to pass, though.

PR #131, opened 2024-03-27
Remove neutron customization from local.conf

vyzigold:
We have an issue in CI, where devstack can't install successfully because of neutron plugin failing to copy some files.

This is an attempt to fix the CI by moving the local.conf a little closer to what worked for me when deploying devstack localy recently.


vyzigold:
This confirms, that my devstack fix works. We just need to wait for it to be merged to unblock the CI.

PR #134, opened 2024-04-29
Bump actions checkout to v4

vkmc:
Bump actions checkout to v4 since Node.js 16 actions are deprecated.
We need to update to Node.js 20, which is included in actions/checkout@v4.

Also bumps setup-go, actions-goveralls and github-script


elfiesmelfie (approved):
I checked the actions, and they're pulling the new versions and running successfully.

csibbitt (approved):

PR #136, opened 2024-06-13
Fix integration CI

vyzigold:
- Disable the logging job. Logging wasn't ever actually used. We can reenable the job if we ever decide it's useful in the future.

- Add some more debug output to other jobs.

- Remove sg-core-ci container from the lint job. Docker isn't able to pull the image because of deprecated docker image manifest. I couldn't find what the container actually does, but as far as I can tell the linter runs the same even without it.


vkmc (approved):
LGTM, thanks Jaromir!

PR #137, opened 2024-06-14
Fix sg-bridge in integration CI

vyzigold:
We were missing a volume in the sg-bridge container and a package, which is required for building sg-bridge from source. I tested with a same named branch in sg-bridge repo, so the CI should be green even on a run after merge.


vkmc (approved):
LGTM, nice debugging work :), thanks!

PR #138, opened 2024-08-05
Update base image to RH UBI 9 and Golang 1.22

vkmc:


vkmc:
recheck

vkmc:
recheck

build/repos/opstools.repo
```diff
@@ -1,19 +1,18 @@
...
 #
-# Please see http://wiki.centos.org/SpecialInterestGroup/OpsTools for more
-# information
+# Please see https://trunk.rdoproject.org/ and
+# https://trunk.rdoproject.org/centos9-caracal/report.html for more information
 
-[centos-opstools-testing]
-name=CentOS-OpsTools - testing repo
-baseurl=https://buildlogs.centos.org/centos/$releasever-stream/opstools/$basearch/collectd-5/
+[delorean-caracal-testing]
```
vyzigold, line 6: Since the file is not about opstools only anymore, maybe it could be renamed? openstack-repos.repo or something like that?

vkmc:
Last missing piece here is coveralls and linting.

Seems coveralls golang plugin hasn't been updated for a year already, and their own repo reflects low coverage. See https://github.com/shogo82148/actions-goveralls. We should look for an alternative or just drop it.

In the case of linting, I had to disable a few linters since the changes required to get it passing might be better done in a follow up patch.

.github/workflows/integration.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         if: steps.bridge_branch.outcome == 'success'
         run: |
           docker run --name=sgbridge --network host $BRIDGE_VOLUME -d -uroot \
-            -e GITHUB_REF -e BRIDGE_SOCKET -e QDR_CHANNEL -e OPSTOOLS_REPO \
+            -e GITHUB_HEAD_REF -e BRIDGE_SOCKET -e QDR_CHANNEL -e OPENSTACK_REPO \
```
csibbitt, line 77: How did you notice this and why does it matter here? I would have though you can pull from either ref for building purposes.

.golangci.yaml
```diff
@@ -21,12 +13,19 @@ issues:
...
+    - plugins/transport/dummy-logs
+    - plugins/application/print
+    - devenv
 
 linters:
   disable-all: true
   enable:
     - bodyclose
-    - depguard
+    # - depguard
```
csibbitt, line 28: Did you look into replacements for any of these disabled linters? Sometimes there is a 1:1 replacement or a spiritual successor mentioned in the release notes.

csibbitt (approved):

PR #143, opened 2024-09-26
Parse custom labels from ceilometer metadata

vyzigold:
related ceilometer patch: https://review.opendev.org/c/openstack/ceilometer/+/926176

Metrics without the user_metadata field are unchanged (current behavior is preserved). When the user_metadata field is present in the metric received from ceilometer it's parsed and each key=value pair is used as a separate metric label.


vyzigold:
recheck

vyzigold:
recheck

elfiesmelfie (approved):

PR #144, opened 2024-10-14
Update base image to RH UBI 9 and Golang 1.21.13

vkmc:


.github/workflows/integration.yml
```diff
@@ -1,7 +1,7 @@
 name: Integration testing
 env:
   PROJECT_ROOT: /root/go/src/github.com/infrawatch/sg-core
-  OPSTOOLS_REPO: https://raw.githubusercontent.com/infrawatch/sg-core/04dcb34edd2c234b378222d2f9a17e15c0dad936/build/repos/opstools.repo
+  OPSTOOLS_REPO: https://trunk.rdoproject.org/centos9-caracal/delorean-deps.repo
```
vkmc, line 4: If we want to avoid enabling unnecessary repos, we would need to update this afterwards to point to the repo file in infrawatch/sg-core

mrunge (approved):
Looks good to me, thank you for working on this!

PR #148, opened 2025-02-14
Update GH actions to use ubuntu-latest

vkmc:
The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025.

Actions need to be updated to a newer version.

For some jobs we are already using ubuntu-latest, so use the same for the jobs using ubuntu20.04


ayefimov-1 (approved):
looks fine.

PR #162, opened 2026-01-21
STF 1.5.7 release ops

vkmc:
- Pin Ceilometer to stable/2024.2
- Bump gopkg.in/yaml.v3 from 3.0.0 to 3.0.1
- Protect the expiryProc list with a mutex
- Fix long message handling with dynamic buffer growth
- Update to Golang 1.24.11
- Update coveralls-badge workflow to be non-blocking


vkmc:
We should fix the test coverage before moving forward with the release.

elfiesmelfie:
recheck

vkmc:
Test coverage was improved, still the reported coverage is lower than before

It is worth noting that the test coverage was dropped once we updated the Golang version from 1.21 to 1.24.11. There might be some differences on how the coverage reports are done.

Added tests for components that were recently updated (expiration mechanism for Prometheus metrics, socket transport). Also added some tests for components that are in use and lacked coverage (parsers for Ceilometer, collectd and Sensubility, and the manager code)

I think we are good to go with the current coverage and, if relevant test coverage is missing, we can add it in a follow up effort.

vkmc:
recheck

vyzigold:
> Test coverage was improved, still the reported coverage is lower than before
> 
> It is worth noting that the test coverage was dropped once we updated the Golang version from 1.21 to 1.24.11. There might be some differences on how the coverage reports are done.
> 
> Added tests for components that were recently updated (expiration mechanism for Prometheus metrics, socket transport). Also added some tests for components that are in use and lacked coverage (parsers for Ceilometer, collectd and Sensubility, and the manager code)
> 
> I think we are good to go with the current coverage and, if relevant test coverage is missing, we can add it in a follow up effort.

Agreed. Fine with me. Big thanks for the new tests!

vyzigold (approved):

PR #166, opened 2026-01-22
Improve test coverage for prometheus application plugin

vkmc:
Add tests for New, Config, PromCollector methods, metricExpiry,
collectorExpiry, ReceiveMetric, UpdateMetrics, Describe, and Collect.

Omit logging and HTTP server tests.

### DIFF
--- a/plugins/application/prometheus/main_test.go
+++ b/plugins/application/prometheus/main_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/infrawatch/apputils/logging"
 	"github.com/infrawatch/sg-core/pkg/data"
 	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -80,11 +81,10 @@ this is: not: valid: yaml
 
 		config := `
 host: 0.0.0.0
-port: 9090
 `
 		err := prom.Config([]byte(config))
 		require.NoError(t, err)
-		assert.Equal(t, 9090, prom.configuration.Port)
+		assert.Equal(t, 3000, prom.configuration.Port)
 	})
 }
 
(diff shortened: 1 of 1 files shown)


plugins/application/prometheus/main_test.go
```diff
@@ -0,0 +1,683 @@
...
+		prom := app.(*Prometheus)
+
+		config := `
+this is: not: valid: yaml
+`
+		err := prom.Config([]byte(config))
+		require.Error(t, err)
+	})
+
+	t.Run("default port when not specified", func(t *testing.T) {
```
vyzigold, line 78: The testcase is called "default port when not specified" - which seems like a good testcase for testing the config. But you're actually specifying the port below, which doesn't seem to make sense to me. Maybe the L83 should be removed?

plugins/application/prometheus/main_test.go
```diff
@@ -0,0 +1,683 @@
...
+		metrics := []prometheus.Metric{}
+		for metric := range ch {
+			metrics = append(metrics, metric)
+		}
+
+		assert.Equal(t, 1, len(metrics))
+	})
+
+	t.Run("collect with timestamp", func(t *testing.T) {
+		pc := NewPromCollector(lw, 1, true)
```
vyzigold, line 448: I'm not sure what's actually the difference between collecting with and without timestamps. Looking at the 2 testcases, the only difference I see is the bool value passed to NewPromCollector.

I wonder if there is some assert that we could do to to assert that the difference in the metrics we expect (if there actually is a difference we expect) is really there? If you think there is nothing more to assert here, then the testcases are fine.

vkmc: For what I understand, in one case we require the collection with a timestamp, so the metric is sent to Prometheus with a timestamp using prometheus.NewMetricWithTimestamp() and in the other case metric is sent to Prometheus without any timestamp (just the raw metric).

We can add assertions to verify this is the behavior.

vyzigold: thank you. I currently see some linter errors, but once the CI is passing, I don't have any further comments and this pr lgtm.

PR #170, opened 2026-04-14
Update to golang 1.25.9

vkmc:
Update to golang 1.25.9 and golangci-lint 2.11.4

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,28 +1,6 @@
 version: 2
 
 issues:
-  exclude-rules:
-    - linters:
-      - errcheck
-      text: "[a-zA-Z]+.[a-zA-Z]+.(Error|Info|Debug|Warn)" # from logger
-    - text: "[A-Z]+" #omit enums
-      linters:
-        - deadcode
-    - text: New
-      linters:
-        - deadcode
-    - linters:
-        - staticcheck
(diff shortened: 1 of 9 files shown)


.golangci.yaml
```diff
@@ -32,28 +83,34 @@ issues:
     - devenv
 linters:
   disable-all: true
+  disable:
```
vkmc, line 14: In golangci-lint v2, there are "Enabled by default linters" (errcheck, govet, ineffassign, staticcheck, unused) that are active even with disable-all: true. So we do need the disable section to override these defaults.

vkmc:
A major update and cleanup is proposed as part of this PR.

We carry over many linting issues since this is already a legacy code base.

Ideally we would address the linting issues, but since infrawatch/sg-core is in maintenance mode and no new code is being added I don't think we should work on fixing those.

vkmc:
recheck

vkmc:
recheck

vkmc:
recheck

vkmc:
recheck

vkmc:
CI is unstable, not related to this change

vkmc:
recheck

vkmc:
recheck

vyzigold (approved):

elfiesmelfie (approved):

PR #179, opened 2026-08-04
Enable TLS certificate verification for Elasticsearch client

danpawlik:
Instead of silently disabling TLS verification when tlsServerName is not provided, derive the server name from the hostURL hostname. This ensures certificate validation is always performed when TLS is enabled, preventing potential MITM attacks on connections carrying credentials and telemetry data.


plugins/application/elasticsearch/pkg/lib/client.go
```diff
@@ -42,16 +43,23 @@ func createTLSConfig(serverName string, certFile string, keyFile string, caFile
 	certPool := x509.NewCertPool()
 	certPool.AppendCertsFromPEM(ca)
 
+	if len(serverName) == 0 {
+		u, err := url.Parse(hostURL)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse hostURL for TLS server name: %w", err)
+		}
+		serverName = u.Hostname()
```
vyzigold: Do you think this is necessary? AFAIK when using TLS without the ServerName configured, the server certificates will automatically be checked against the URL used to send requests, so I think you just reimplemented the default behavior here.

ServerName would typically be used in cases where the dnsNames in the cert don't match the URL used for accessing the server. For example when you have an elasticsearch configured with certificates which say its DNS is "es.stf.svc", but you configure sg-core to access it through its IP address. Then you'd set the ServerName to "es.stf.svc". In these cases I'd expect the serverName to be included in the config file.

danpawlik: make sense

